Issue #10, opened 2020-06-13
would you be ok with setvalue(uint8_t& data8)?


mitchjs:
i like to have this, just so i dont have to use the one with length for a simple byte
i mean we got others (16,32...)
and it just looks nicer in user code

pCharacteristic->setValue(btBypassEnabled); 

in BLECharacteristic.cpp

```
void BLECharacteristic::setValue(uint8_t& data8) {
	uint8_t temp = uint8;
	setValue((uint8_t*)&temp, 1);
} // setValue
```

thanks
mitch

h2zero:
Yep, that's certainly a good addition.

mitchjs:
if i knew how exactly to use GIT i would do it for ya... but its so tiny... prob not worth me doing... you can paste that in and add the method to .h file
(other than git clone and git pull :) )

h2zero:
No worries, I'll take care of it 😄

mitchjs:
reminder....

h2zero:
done

mitchjs:
thanks

PR #14, opened 2020-06-19
Don't send notifications to un-paired peers when READ_ENC set

h2zero:
If read encryption is applied to a characteristic the server should not allow notifications of it's value to be sent to peers that have not paired with the server.

* Also added minor bugfixes

### DIFF
--- a/src/NimBLECharacteristic.cpp
+++ b/src/NimBLECharacteristic.cpp
@@ -137,7 +137,7 @@ uint16_t NimBLECharacteristic::getHandle() {
 } // getHandle
 
 
-uint8_t NimBLECharacteristic::getProperties() {
+uint16_t NimBLECharacteristic::getProperties() {
     return m_properties;
 } // getProperties
 
--- a/src/NimBLECharacteristic.cpp
+++ b/src/NimBLECharacteristic.cpp
@@ -346,6 +346,9 @@ void NimBLECharacteristic::notify(bool is_notification) {
 
     std::string value = getValue();
     size_t length = value.length();
+    bool reqSec = (m_properties & BLE_GATT_CHR_F_READ_AUTHEN) ||
+                  (m_properties & BLE_GATT_CHR_F_READ_AUTHOR) ||
+                  (m_properties & BLE_GATT_CHR_F_READ_ENC);
     int rc = 0;
 
     for (auto &it : p2902->m_subscribedVec) {
@@ -353,8 +356,16 @@ void NimBLECharacteristic::notify(bool is_notification) {
 
         // check if connected and subscribed
         if(_mtu == 0 || it.sub_val == 0) {
-            //NIMBLE_LOGD(LOG_TAG, "peer not connected");
             continue;
+        }
+
+        // check if security requirements are satisfied
+        if(reqSec) {
+            struct ble_gap_conn_desc desc;
+            rc = ble_gap_conn_find(it.conn_id, &desc);
+            if(rc != 0 || !desc.sec_state.encrypted) {
+                continue;
+            }
         }
 
         if (length > _mtu - 3) {
--- a/src/NimBLECharacteristic.h
+++ b/src/NimBLECharacteristic.h
@@ -107,7 +107,7 @@ private:
     ~NimBLECharacteristic();
 
     NimBLEService*  getService();
-    uint8_t         getProperties();
+    uint16_t        getProperties();
     void            setSubscribe(struct ble_gap_event *event);
     static int      handleGapEvent(uint16_t conn_handle, uint16_t attr_handle,
                                    struct ble_gatt_access_ctxt *ctxt, void *arg);
--- a/src/NimBLEServer.cpp
+++ b/src/NimBLEServer.cpp
@@ -230,7 +230,6 @@ size_t NimBLEServer::getConnectedCount() {
             else {
                 server->m_connectedPeersVec.push_back(event->connect.conn_handle);
 
-                ble_gap_conn_desc desc;
                 rc = ble_gap_conn_find(event->connect.conn_handle, &desc);
                 assert(rc == 0);
 
@@ -276,6 +275,17 @@ size_t NimBLEServer::getConnectedCount() {
 
             for(auto &it : server->m_notifyChrVec) {
                 if(it->getHandle() == event->subscribe.attr_handle) {
+                    if((it->getProperties() & BLE_GATT_CHR_F_READ_AUTHEN) ||
+                       (it->getProperties() & BLE_GATT_CHR_F_READ_ENC))
+                    {
+                        rc = ble_gap_conn_find(event->subscribe.conn_handle, &desc);
+                        assert(rc == 0);
+
+                        if(!desc.sec_state.encrypted) {
+                            NimBLEDevice::startSecurity(event->subscribe.conn_handle);
+                        }
+                    }
+
                     it->setSubscribe(event);
                     break;
                 }
--- a/src/NimBLEServer.cpp
+++ b/src/NimBLEServer.cpp
@@ -181,8 +181,8 @@ int NimBLEServer::disconnect(uint16_t connId, uint8_t reason) {
                                     NimBLEUtils::returnCodeToString(rc));
     }
 
-    return rc;
     NIMBLE_LOGD(LOG_TAG, "<< disconnect()");
+    return rc;
 } // disconnect
 
 
@@ -276,6 +276,7 @@ size_t NimBLEServer::getConnectedCount() {
             for(auto &it : server->m_notifyChrVec) {
                 if(it->getHandle() == event->subscribe.attr_handle) {
                     if((it->getProperties() & BLE_GATT_CHR_F_READ_AUTHEN) ||
+                       (it->getProperties() & BLE_GATT_CHR_F_READ_AUTHOR) ||
                        (it->getProperties() & BLE_GATT_CHR_F_READ_ENC))
                     {
                         rc = ble_gap_conn_find(event->subscribe.conn_handle, &desc);
@@ -340,7 +341,7 @@ size_t NimBLEServer::getConnectedCount() {
         } // BLE_GAP_EVENT_REPEAT_PAIRING
 
         case BLE_GAP_EVENT_ENC_CHANGE: {
-            rc = ble_gap_conn_find(event->conn_update.conn_handle, &desc);
+            rc = ble_gap_conn_find(event->enc_change.conn_handle, &desc);
             if(rc != 0) {
                 return BLE_ATT_ERR_INVALID_HANDLE;
             }


mitchjs:
looking at the changes.... 
trying to learn :)
so... if one connects to device, and trys to ENABLE notifications, will it prompt for pairing?
i guess i need to try this PR and see...
i will tomorrow :) late now so...sleepy...

thanks @h2zero !

h2zero:
@mitchjs yes it will prompt for pairing, unfortunately because of how NimBLE handles the descriptor it will not notify the client in the case of failure. Instead what I have done is check for the pairing and if not paired we will not send the value to the client. I recommend using the callback for this to disconnect like so:
```
void onAuthenticationComplete(ble_gap_conn_desc* desc) { 
    if(!desc->sec_state.encrypted) {
        NimBLEDevice::getServer()->disconnect(desc->conn_handle);
    }
}
```

mitchjs:
ok.. so i didnt add that callback (YET)

i got the pair request, put in a invalid pin, 111111
and then turned on notifications, got another pair request
said pair, it never prompted for pin
but it is returning valid data....

and in IOS bluetooth devices, its listed like it bonded....


something is really wrong

mitchjs:
it seems that second pair request, where i dont enter any pin, causes a BOND
and thats a big secuirty hole
so basicly i got a bond, with an INVALID PIN

i hope im doing this right

 **NimBLEDevice::setSecurityIOCap(BLE_HS_IO_DISPLAY_ONLY);
    NimBLEDevice::setSecurityAuth(BLE_SM_PAIR_AUTHREQ_BOND);
    NimBLEDevice::setSecurityPasskey(123456);**

and got READ_ENC and WRITE_ENC used in my characteristics

mitchjs:
so can one bond with invalid pin??!?!?!?!
when i enable the notifications in lightblue, the popup asking to pair, shows
(but no pin requested, i already put in an invalid pin)
i say pair...

I (150872) DS-EZLOCK-ECM9(BLE): BT - onAuthenticationComplete() - ENCRYPTED
I (150872) DS-EZLOCK-ECM9(BLE): BT - onAuthenticationComplete() - BONDED
I (150872) DS-EZLOCK-ECM9(BLE): BT - onAuthenticationComplete() keysize=16

but not AUTHENTICATED....

but worse all characteristics are accessible, like there is no security at all

mitchjs:
this could be all me, not fully understanding BLE security :)
and maybe i want

READ_AUTHEN over READ_ENC :)

h2zero:
I’m not quite sure what’s happening for you here. I can tell you for sure that if you are bonded with the esp then you will always have an encrypted/authenticated connection. 

If you erase bonds from both devices and try again is the problem repeated?

mitchjs:
yes... it seems an invalid pin, still eventually bonds/encrypts
i would think a bad pin = you cant get any data and wont be bonded

mitchjs:
only a valid pin, gets me authentication

mitchjs:
its worse, i enter wrong pin, then enable notify for 1 of my chars, lightblue pops up a request to pair (no pin entry) i press pair, now notifications work for this char
its bonded
so i disconnect
reconnect, get prompted for PIN, hit cancel...
disconnect, now cant reconnect... it connects, then disconnects

mitchjs:
if i just disconnect when bad pin entered... it should be ok... but is that how BT things work?
most of my BT devices dont need any auth, but i need to make sure no one messes with this product
unless they should, ie know the passkey to pair/bond

mitchjs:
another wierd thing, if i turn on other notifications on other chars, they dont work
only that 1st one i enable .... 

OK... so i did

```
if(!desc->sec_state.authenticated)
{
    ESP_LOGI(LOG_TAG, "BT - onAuthenticationComplete() - NOT AUTHENTICATED, DISCONNECTING...");
            
    NimBLEDevice::getServer()->disconnect(desc->conn_handle);
}
```

and sure, it does work... if bad pin, connection dropped

h2zero:
Strange indeed. I will do some more testing after work and see what I can find.

mitchjs:
thanks...

so i been testing with iphone/lightblue
i grabbed my android tablet, and it fails to connect! (using lightblue)
(i turned off the security in code... )
well it connects, but fails to get anything from device... services
going to go back to my simple test code
yikes!
my simple 1 characteristic test, no security... no working on android tablet at all
both in lightblue and nRF Connect... 
it just wont fully connect and show the service with the 1 char

what do you test with? ios or android?

h2zero:
Wow, all kinds of issues I haven’t seen lol. I test with other esp32s mostly and iOS, don’t have my android tablet anymore so I’ll see if I can borrow one to test with.

mitchjs:
**Panic** time, my **bluedroid** version(of my full product), works just fine on my Android tablet

h2zero:
Found some interesting behavior when I enabled NIMBLE_DEBUG. I'll need to wait until I'm home to sort it out.

mitchjs:
FYI - bluedroid

with secuirty enabled
when i attempt to turn on a Notification, i get error in light blue saying failed to enable
and in log of esp32
E (14422) BT_GATT: gatts_write_attr_perm_check - GATT_INSUF_ENCRYPTION
which i like, cause no pin/bond has been done yet
also no request for pin yet...
now i do read , i get the prompt for pin
bad pin, does disconnect me :)
good pin, and its all bonded and ok

mitchjs:
ok, i made a simple, 1 service, 1 char with read_enc and notify_enc
using bluedroid
and it works as expected on android
the nimble doesnt seem to work at all on android (without security)
attached is the bluedroid main.cpp
if you wanted to test  and compare

[main.cpp.txt](https://github.com/h2zero/esp-nimble-cpp/files/4806888/main.cpp.txt)



if you need a tablet, i can send you this tablet

h2zero:
> the nimble doesnt seem to work at all on android (without security)

Are you saying nimble is working on your tablet with security enabled? If so I think the reason would be due to already being bonded.

mitchjs:
no nimbled doesnt seem to work at all!
no security...

mitchjs:
one thing i just noticed

the bluedroid creates 1 more characteristic on the GENERAL ACCESS Service
the "CentralAddressResolution" 

Bluetooth.com says its requirement Mandatory
[https://www.bluetooth.com/xml-viewer/?src=https://www.bluetooth.com/wp-content/uploads/Sitecore-Media-Library/Gatt/Xml/Characteristics/org.bluetooth.characteristic.gap.central_address_resolution.xml](url)

mitchjs:
seems its an option in nimble

BLE_SVC_GAP_CENTRAL_ADDRESS_RESOLUTION needs to be 1

now to find where to set that

h2zero:
Thanks, I’m going to do some testing as there may be an undetected bug in the server refactor code I recently pushed. Also not too sure about the controller in the idf version you’re using, if you could test with idf master and see if the problem continues that would be great. 

You’re using idf 4.0.1 correct?

mitchjs:
yes... 4.0.1 but nimble is 1.2.0-idf
 do you know where to set nimble config
i want to try BLE_SVC_GAP_CENTRAL_ADDRESS_RESOLUTION 1
then it looks like it will add this char

mitchjs:
i switched to master, NO GO
id like to see if we can get BLE_SVC_GAP_CENTRAL_ADDRESS_RESOLUTION 
working... its a setting of nimble
you can see in
\esp-idf-master\components\bt\host\nimble\nimble\nimble\host\services\gap\src\ble_svc_gap.c

h2zero:
Yeah, that won't work because the ESP32 uses it's own proprietary controller and nimble has it's own that we aren't using. Welcome to the rabbit hole lol, bonding and resolving phone mac addresses is done on the host side of the ESP32, which is what we are working with so it gets confusing. 

I have my testing setup the same as yours currently, I hope I can reproduce the problems.

mitchjs:
so how do i turn on nimble debug?

mitchjs:
well it did enable that central address char, but it didnt help the android connection...
oddly windows 10, Bluetooth LE Explorer... is happy with nimble either way
it just seems to HANG when android connects
i get the callback onConnect()
and thats it

h2zero:
Yeah I'm not having any trouble here, using the test code you've provided (also adapted the file you posted) I cannot reproduce so far. 

NimBLE debug; menuconfig > component config > bluetooth > nimble options >enable extra runtime asserts and host debugging

mitchjs:
i did that check box, and rebuilt i noticed no new output in esp32 logs

im going to buy a new android tablet... this one is stuck on 5.1 lollypop...
ps. i hate android.... but the app to control my device, is coded for it :)

mitchjs:
ok, got debug on....


D NimBLEServerCallbacks: "onConnect(): Default"
ble_hs_hci_cmd_send: ogf=0x08 ocf=0x0016 len=2
0x16 0x20 0x02 0x00 0x00
Command Status: status=0 cmd_pkts=5 ocf=0x16 ogf=0x8
LE Remote Used Features. FAIL (status=34)
Disconnection Complete: status=0 handle=0 reason=34
D NimBLEServer: ">> handleGapEvent: "
I (87516) MyBLEServerCallbacks: onDisconnect()

im sure that not helpful

h2zero:
It's actually very helpful;
`LE Remote Used Features. FAIL (status=34)`
means:
0x22  BLE_ERR_LMP_LL_RSP_TMO  LMP Response Timeout / LL Response Timeout

The question now is why, possibly an android issue, possibly nimble, needs more investigation. Thanks!

mitchjs:
this is old

https://www.mail-archive.com/dev@mynewt.incubator.apache.org/msg03494.html

mitchjs:
DING DING DING

i set MYNEWT_VAL_BLE_LL_CFG_FEAT_DATA_LEN_EXT to 0
in: esp-idf-master\components\bt\host\nimble\port\include\esp_nimble_cfg.h

connects, and show services :)

h2zero:
That was some quick deduction, well done 👍. I will have to look into this, possibly a PR for the upstream code.

mitchjs:
well....
i sent my code to a friend, who has newer tablet, android 8
and it just worked...

so.. that message is from 2017, it sounds like they would have fixed it by now
not sure where there esp32 version is from(date wise)

and whats the downside of

MYNEWT_VAL_BLE_LL_CFG_FEAT_DATA_LEN_EXT (0)

mitchjs:
here is what it is

Setting: BLE_LL_CFG_FEAT_DATA_LEN_EXT
    * Description: This option enables/disables the data length update procedure in the controller. If enabled, the controller is allowed to change the size of tx/rx pdu's used in a connection.

h2zero:
I'm not 100% sure but I believe it to be that it allows the devices to exchange MTU size to accommodate larger amounts of data in a single read/write operation.

h2zero:
lol good work again, so basically your android tablet does not support that feature (bluetooth 4.0) where nimble is looking for 4.1, or something to that effect. So that means the downside of changing that value would be the lack of capability to perform MTU upgrades, at least that's my interpretation.

mitchjs:
yea... my guess is one should ask the device if it supports this.. and they dont
or at least not in 1.2.0
there is newer, but the espessif not touched it

h2zero:
I saw a PR related to this, just found it [here](https://github.com/apache/mynewt-nimble/pull/781) maybe give that a try?

mitchjs:
it refers to PR782, and that just removes one line of code...
and i looked at the esp32 nimble and that line is not there already
so thats not it

h2zero:
You're right, also I was just testing the capability of MTU exchanging with BLE_LL_CFG_FEAT_DATA_LEN_EXT set to 0 and saw no difference in effect. I'm not really sure what it's purpose is, I'll need to read the bluetooth specs on it.

h2zero:
Well, I found a better explanation [here](https://punchthrough.com/maximizing-ble-throughput-part-3-data-length-extension-dle-2/) basically bluetooth 4.2 is required as a minimum for both devices for that functionality to work, I'm guessing your tablet does not support that.

mitchjs:
right, and my device should not be reporting that it can...
i see the code, where the feature request is made, and response is processed
and where a bit is checked if it supports DLE

mitchjs:
looks like this was a wild goose chase, none of the above mattered
in fact none of that code that uses BLE_LL_CFG_FEAT_DATA_LEN_EXT get compiled
files in esp-idf-v4.0.1\components\bt\host\nimble\nimble\nimble\controller
dont get use, i think the are just reference files

it all started working, when log debug on...
ill i have to do is set menuconfig to log level debug
and it works

put it back to INFO, and it doesnt

when it works:
(the error is BLE_ERR_UNSUPP_REM_FEATURE)
```
LE connection complete. handle=0 role=1 paddrtype=1 addr=d0.1d.c1.5a.d6.76 local_rpa=0.0.0.0.0.0 peer_rpa=0.0.0.0.0.0 itvl=30 latency=0 spvn_tmo=2000 mca=2
D NimBLEServer: ">> handleGapEvent: "
I (312303) DS-EZLOCK-ECM9(BLE): BT - connected
D NimBLEServerCallbacks: "onConnect(): Default"
ble_hs_hci_cmd_send: ogf=0x08 ocf=0x0016 len=2
0x16 0x20 0x02 0x00 0x00
Command Status: status=0 cmd_pkts=5 ocf=0x16 ogf=0x8
LE Remote Used Features. FAIL (status=26)
ble_hs_hci_evt_acl_process(): conn_handle=0 pb=2 len=11 data=0x07 0x00 0x04 0x00 0x10 0x01 0x00 0xff 0xff 0x00 0x28 
rxed att command: read group type req; conn=0 start_handle=0x0001 end_handle=0xffff
txed att command: read group type rsp; conn=0 length=6
host tx hci data; handle=0 length=24
ble_hs_hci_acl_tx(): 0x00 0x00 0x18 0x00 0x14 0x00 0x04 0x00 0x11 0x06 0x01 0x00 0x05 0x00 0x00 0x18 0x06 0x00 0x09 0x00 0x01 0x18 0x0a 0x00 0x10 0x00 0x0a 0x18 
Number of Completed Packets: num_handles=1
handle:0 pkts:1
ble_hs_hci_evt_acl_process(): conn_handle=0 pb=2 len=11 data=0x07 0x00 0x04 0x00 0x10 0x11 0x00 0xff 0xff 0x00 0x28 
rxed att command: read group type req; conn=0 start_handle=0x0011 end_handle=0xffff
txed att command: read group type rsp; conn=0 length=20
host tx hci data; handle=0 length=26
```

how odd is that???
for sure with log level set to info, it hangs... then timesout

mitchjs:
it still fails sometimes...
a fail is:

LE connection complete. handle=0 role=1 paddrtype=1 addr=d0.1d.c1.5a.d6.76 local_rpa=0.0.0.0.0.0 peer_rpa=0.0.0.0.0.0 itvl=30 latency=0 spvn_tmo=2000 mca=2
D NimBLEServer: ">> handleGapEvent: "
I (184803) DS-EZLOCK-ECM9(BLE): BT - connected
D NimBLEServerCallbacks: "onConnect(): Default"
ble_hs_hci_cmd_send: ogf=0x08 ocf=0x0016 len=2
0x16 0x20 0x02 0x00 0x00 
Command Status: status=0 cmd_pkts=5 ocf=0x16 ogf=0x8

success is:

LE connection complete. handle=0 role=1 paddrtype=1 addr=d0.1d.c1.5a.d6.76 local_rpa=0.0.0.0.0.0 peer_rpa=0.0.0.0.0.0 itvl=30 latency=0 spvn_tmo=2000 mca=2
D NimBLEServer: ">> handleGapEvent: "
I (230403) DS-EZLOCK-ECM9(BLE): BT - connected
D NimBLEServerCallbacks: "onConnect(): Default"
ble_hs_hci_cmd_send: ogf=0x08 ocf=0x0016 len=2
0x16 0x20 0x02 0x00 0x00
Command Status: status=0 cmd_pkts=5 ocf=0x16 ogf=0x8
LE Remote Used Features. FAIL (status=26)
ble_hs_hci_evt_acl_process(): conn_handle=0 pb=2 len=11 data=0x07 0x00 0x04 0x00 0x10 0x01 0x00 0xff 0xff 0x00 0x28 
rxed att command: read group type req; conn=0 start_handle=0x0001 end_handle=0xffff
txed att command: read group type rsp; conn=0 length=6
host tx hci data; handle=0 length=24
ble_hs_hci_acl_tx(): 0x00 0x00 0x18 0x00 0x14 0x00 0x04 0x00 0x11 0x06 0x01 0x00 0x05 0x00 0x00 0x18 0x06 0x00 0x09 0x00 0x01 0x18 0x0a 0x00 0x10 0x00 0x0a 0x18
Number of Completed Packets: num_handles=1
...

is there anything in the the c++ code that could be hanging? some task notify? some semaphore?

h2zero:
Try commenting out [this line](https://github.com/espressif/esp-nimble/blob/fead24e5d5c0f4bd9b8c3d71cf72a87d75631399/nimble/host/src/ble_gap.c#L1691).

I can't think of anywhere the C++ code would hang, but I have seen similar behavior on occasion but in different code areas (start advertising for instance) hard to reproduce and have not encountered it lately.

mitchjs:
ok, i commented out that line... and it connected and does not hang
confirmed, uncomment, and try it hangs

so how did you know about that... and what to do about it
what does that function do?

h2zero:
The error code you are getting is handled [here](https://github.com/apache/mynewt-nimble/blob/2da39a1fe53a635252cc04cac3f8df7305af4eca/nimble/host/src/ble_gap.c#L1187) it looks like that is where the timeout is coming from. When the operation fails it tries to perform it over l2cap and sets a timer. This seems to be that the device you are connecting to does not support connection parameter updates, at least that's my first coffee of the day assessment.

> so how did you know about that... and what to do about it
what does that function do?

I've been in there troubleshooting before, it requests the features of the peer.

mitchjs:
ok.. so it fails... and say device doesnt support connection parameter updates... thats fine too
so why hang...

h2zero:
I'm not sure, if you turn on the event codes to strings option and watch the log for `BLE_GAP_EVENT_L2CAP_UPDATE_REQ` that might be the issue (server doesn't have a handler for that)

mitchjs:
for sure hangs: Command Status: status=0 cmd_pkts=5 ocf=0x16 ogf=0x8
so thats the command 0x16 going out

but then it times out, and i do it again and it works
but only when DEBUG is on

D NimBLEServerCallbacks: "onConnect(): Default"
ble_hs_hci_cmd_send: ogf=0x08 ocf=0x0016 len=2
0x16 0x20 0x02 0x00 0x00
Command Status: status=0 cmd_pkts=5 ocf=0x16 ogf=0x8
LE Remote Used Features. FAIL (status=26)

h2zero:
It would seem it hangs on some concurrent access that only happens with minimal debug logs, where it happens is a mystery... if only I could reproduce this. There are locks and semaphores used in the nimble stack code where it might hang (in my experience most issues in there are the code espressif added).

h2zero:
[Here](https://github.com/espressif/esp-idf/blob/7be7839690f814d0a6aa5d4dd106dd43aec80354/components/bt/host/nimble/esp-hci/src/esp_nimble_hci.c#L82) is one that I've dealt with before. Wasn't the issue, just an example. It's very hard to trace through and at that point of the code it goes to the controller.

mitchjs:
see my update, i never works with debug off... 
so timing?

mitchjs:
when it fails it never gets to ble_hs_hci_evt_le_rd_rem_used_feat_complete()
seems it doesnt FAIL, and hangs... need it to get that **LE Remote Used Features. FAIL (status=26)**
as after that it moves on

mitchjs:
ok.. IPHONE

ble_hs_hci_cmd_send: ogf=0x08 ocf=0x0016 len=2
0x16 0x20 0x02 0x00 0x00 
Command Status: status=0 cmd_pkts=5 ocf=0x16 ogf=0x8
**LE Remote Used Features. handle=0 feat=ff 79 01 00 00 00 00 00**


so its clear, that the android(junk) samsung tablet doesnt support this at BT 4.0
and nimble, not always getting to the FAIL...

mitchjs:
ps. i did order a brand new (well 2019 samsung tab 8)
just to try this on...
then ill sell the old one on ebay... and call it a day
odd that bluedroid doesnt have this issue...
im guessing some timeout in nimble not happening?
question is where in code should it all be handled...
is it all done in the low level ESP32 code... HCI stuff

h2zero:
That seems to be the case, something is hanging somewhere for a response that doesn’t come. I’ll see what I can find when I have some more time.

h2zero:
I have a theory if you wouldn't mind testing?
Add `vTaskDelay(1 / portTICK_PERIOD_MS);` [here](https://github.com/h2zero/esp-nimble-cpp/blob/afb2f8d4c7b02ddb23c5397526dd73865caa191e/src/NimBLEServer.cpp#L240)

That should force a task switch on cpu1 and allow the controller to do whatever it's doing before the call to get the peer features. I think the controller is getting stuffed up.

h2zero:
Alternatively, as a test, you could add `esp_bt_sleep_disable();` [here](https://github.com/h2zero/esp-nimble-cpp/blob/afb2f8d4c7b02ddb23c5397526dd73865caa191e/src/NimBLEDevice.cpp#L464) as the issue could also be related to the controller going to sleep.

h2zero:
@mitchjs we got a fair bit off topic here, I would like to get back to the start of the issue here

> i got the pair request, put in a invalid pin, 111111
and then turned on notifications, got another pair request
said pair, it never prompted for pin
but it is returning valid data.... (notifications are happening)

and in IOS bluetooth devices, its listed like it bonded....

I cannot reproduce this, I have a characteristic with READ_ENC and when I try to read it or subscribe to notifications it prompts for the pin, if I enter incorrectly I'm unable to read the value and no notifications are sent. Because nimble doesn't have WRITE_ENC on the descriptor it allows nrfconnect to believe it is subscribed and my workaround just never sends values to it.

When bonding is enabled I cannot bond until I enter the correct pin, but once I do I am not asked for the pin ever again and can freely read/subscribe anything.

mitchjs:
we sure did get far off topic :) (i will try those things above)

if i remember, with bluedroid, when i turned on notifications, i didnt get prompted for PIN
it just failed

but i think it was more than READ_ENC needed on nimble, it was READ_AUTHEN that is needed
cause you can be encrypted but not authorized

i will review on ios again, and reshen my memory on the topic

h2zero:
> if i remember, with bluedroid, when i turned on notifications, i didnt get prompted for PIN
it just failed

This would be ideal, unfortunately it's not something we can change here. I will be bringing this issue up in the upstream repo, but I'd like to get this as good as I can first (a PR there could take months to get included in IDF)

> but i think it was more than READ_ENC needed on nimble, it was READ_AUTHEN that is needed
cause you can be encrypted but not authorized

That would be odd and I don't see that here but if you can confirm and possibly provide an example I'd like to look into this.

mitchjs:
i think without AUTHEN, i could BOND with an INVALID PIN
ok.. here is what to do in lightblue on ios...
go to the characteristic, it will popup asking for pin, put in a bad pin...
then at this point...
click on start listening, i get a popup that just says PAIR or CANCEL (no pin to put in)
say PAIR
at this point, it bonds

I (135012) DS-EZLOCK-ECM9(BLE): BT - onAuthenticationComplete() - ENCRYPTED
I (135012) DS-EZLOCK-ECM9(BLE): BT - onAuthenticationComplete() - BONDED

and i can get values back from the notification, and if you disconnect and reconnect
it will be locked out and just connect then drop

I (404422) DS-EZLOCK-ECM9(BLE): BT - connected
I (404652) DS-EZLOCK-ECM9(BLE): BT - disconnected

and you have to delete the bonding in ios

mitchjs:
oh my solution is force a disconnect in callback onAuthenticationComplete() when not authenticated
so there is no chance of bond

maybe the pairing request when turning on the notification is wrong, you added that right?

mitchjs:
here is what i get after i press "listen for notifications"
(and thats after a invalid pin entry from a previous popup)
and if i pair it, it bonds... all with the bad pin... and thats when things go wacko
Screenshot 2020-06-21 at 2 07 54 PM

mitchjs:
bluedroid test comparison
I (6142) TEST-BLUDROID: BT - connected
E (7012) BT_GATT: GATT_INSUF_ENCRYPTION
I (7072) BLEDevice: ESP_GAP_BLE_SEC_REQ_EVT
I (7082) BLEDevice: ESP_GAP_BLE_PASSKEY_NOTIF_EVT
I (7082) BLEDevice: passKey = 123456
**ENTERED INVALID PIN**
I (19132) BLEDevice: ESP_GAP_BLE_AUTH_CMPL_EVT
E (19162) BT_BTM: BTM_GetSecurityFlags false
E (19222) BT_BTM: BTM_GetSecurityFlags false
E (21352) BT_BTM: BTM_GetSecurityFlags false
**ACCESSED CHARACTERISTIC (prob read attempt)**
E (56372) BT_BTM: BTM_GetSecurityFlags false
E (56382) BT_GATT: GATT_INSUF_AUTHENTICATION
**WHEN TRYING TO TURN ON NOTIFICATION:**
E (32512) BT_BTM: BTM_GetSecurityFlags false
E (32512) BT_GATT: gatts_write_attr_perm_check - GATT_INSUF_AUTHENTICATION

h2zero:
Thanks, I’ll try to reproduce and see what’s happening. It should be asking for the pin again when enabling notifications, I suspect the call to start security is failing but notifications and bonding still should not happen.

Also when the Pop up that just has the option to pair or not comes up that’s what happens when using “just works” pairing. So somewhere in your code may be setting it back to that.

mitchjs:
great, here is a test main.cpp that exhibits it
[main.cpp.txt](https://github.com/h2zero/esp-nimble-cpp/files/4810128/main.cpp.txt)

h2zero:
Thanks, going to test now. I already see a possible problem though
```
NimBLEDevice::setSecurityAuth(true, false, false);
```
should be:
```
NimBLEDevice::setSecurityAuth(true, true, true);
```
MITM protection enable because you are using a passkey, that is the protection. Secure connections because it really has no reason to ever be off lol, defaults to legacy anyway if peer does not support it.

h2zero:
Just ran your test code and I reproduced the issue. I resolved it using the above stated settings  `setSecurityAuth(true,true,true)`.

mitchjs:
true, true, true did make a difference... when i tap on start listening.... i get the proper dialog asking for a pin now

i wish it wouldnt turn on notifications.... could be nimble thing.... they shouldnt turn on unless autenticated

more testing to do shortly...

h2zero:
> i wish it wouldnt turn on notifications.... could be nimble thing....

Yes it's a nimble thing and also an app thing, some will write the descriptor without response and just assume it worked, and nimble actually does set the subscription value internally but the library will not send the values to an insecure client if security is set.

mitchjs:
ok, good news
with NimBLEDevice::setSecurityAuth(true, true, true);

and turning on notifications(ugh that should not be allowed)
so putting in bad pin, now NO Notifications come through
and MOST importantly no BAD BONDS

mitchjs:
could got me thinking, why have the options of false...
i wonder if we should remove the forced prompt for security when turning on notification
since it turns on anyway... what is the right thing to do
(besides get it fixed that it shouldnt turn on the notification at all unless connection security is good)

h2zero:
> could got me thinking, why have the options of false...

Yes I wondered this myself, MITM should just set itself based on the the security IOcaps and secure connections should always be on IMO but that's how they designed NimBLE. "Might as well leave the options available" seems to be the design (except in the case of 2902 descriptors lol).

> i wonder if we should remove the forced prompt for security when turning on notification
since it turns on anyway... what is the right thing to do

I put that in there because there is a possibility to have a characteristic with only READ_ENC and NOTIFY enabled so it would trigger the security response and allow that configuration to work.

I'm glad this is working for you now, I will merge this unless there is any other issues with it. This is hopefully a short term workaround and I will raise the issue upstream.

mitchjs:
yea.. go ahead and merge (the changes are really just to the BLE_GAP_EVENT_SUBSCRIBE in NimBLEServer.cpp and if something better comes along we can always tweak from there
can you also commit a change for the setval in #10 

really wish i knew how the notifications should work... should they not enable at all if not secure (thats what bluedroid does, and makes sense)

h2zero:
> can you also commit a change for the setval in #10

Yes I will add that right away as well.

> really wish i knew how the notifications should work... should they not enable at all if not secure (thats what bluedroid does, and makes sense)

What should be done is WRITE_ENC should be enabled on the descriptor, that's what your bluedroid code does essentially. That's what I hope can be accomplished upstream in the future.

h2zero:
BTW we broke a record for me, highest number of comments on a PR ever! 😄

mitchjs:
Could I find where in the nimble code the descriptor is created and hard code in WRITE_ENC?

mitchjs:
Haha well the side track of nimble hanging did it

h2zero:
> Could I find where in the nimble code the descriptor is created and hard code in WRITE_ENC?

Yes you could change `BLE_ATT_F_WRITE` to `BLE_ATT_F_WRITE_ENC` at [this line](https://github.com/apache/mynewt-nimble/blob/cf8e8125fd2428adffd78ef0ea03fdae4faec88c/nimble/host/src/ble_gatts.c#L794)

> Haha well the side track of nimble hanging did it

Did either of my proposed fixes for that a few posts back help with that situation?

mitchjs:
Sadly no... tried the task delay and the sleep disable

h2zero:
Too bad, I'd really like to understand what that issue was, but I cannot reproduce it. Maybe I'll look for an old android tablet.

PR #16, opened 2020-07-02
Fix comments for doxygen

h2zero:
Setting up for document generation.


mitchjs:
is the freeRTOS.cpp/.h file still used by the lib? if so maybe better called NimBLEFreeRTOS.cpp?

h2zero:
Actually it is no longer usedin the library. I've left it so it's available to users so if they used that class in their app code they wouldn't have to change it. Only thing I've changed in the file is some comments.

PR #21, opened 2020-07-22
Server: Add onSubscribe() callback for characteristics.

h2zero:
Adds a new method to NimBLECharacteristicCallbacks that gets called when a client
changes it's subscription status.

* Remove NimBLE2902 class.

As the NimBLE2902 class usefulness was only related to callback functions that were replaced
by the NimBLECharacteristicCallbacks:onSubscribe() method this removes the NimBLE2902 class and
moves all subscription handling to NimBLECharacteristic.

* Update documents to reflect this change.


mitchjs:
git confuses the heck out of me... how do i get this to pull
i did
git switch feature/onSubscribe-callback
git pull

and it doesnt pull...

mitchjs:
need to remove:  "src/NimBLE2902.cpp" from MakeLists.txt

h2zero:
Oops, yes I do need to remove that.

Git fetch
Git checkout -b origin feature/onSubscribe-callback

That should do it.

mitchjs:
i did some testing... everything still works (and the new OnSubscribe() works)
i would say this can be merged to master
surprising amount of code change in NimBLECharacteristic.cpp to eliminate the 2902 class
do we need a isSubscribed? sure can use the getSubscribedCount() to know :) if one needed to know

h2zero:
Thanks @mitchjs, yes I had to do a lot of things to support that class, which in then seems wasted lol. Now if only I could get rid of 2904 too...

IsSubscribed() could likely be removed as well, thanks for pointing it out.

PR #32, opened 2020-09-11
Add duration and callback parameter to NimBLEAdvertising::start

h2zero:
* Adds functionality to advertise for a set duration, similar to NimBLEScan::start

* Callback parameter also added that is invoked when advertising stops.

* New NimBLEAdvertising::isAdvertising method added to report if advertising is currently active.


mitchjs:
is this so you could say, advertise for 30 seconds, and it would stop...
vs managing the timer and starting and stopping yourself?

h2zero:
Hi @mitchjs, yes that is the intent. You can also register a callback for when it ends or poll `isAdvertising()` to know when it stops.

PR #38, opened 2021-03-03
Used more stable arduino detection methods

AronRubin:
I have a project that includes both Arduino as a component and esp-nimble-cpp in an ESP-IDF v4.4 environment. There were two issues preventing that from working. First the ARDUINO_ARCH_ESP32 macro is defined on the arduino component compilation command line and seems not to carry over to other components reliably. The second is the conditional PRIV_REQUIRES of arduino in CMakeLists.txt failed to account for the IDF's dependency gathering phase of configuring. I have remedied both issues. Additionally, I made the CMakeLists.txt more current. Sorry about any style violations in that last step.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,14 +3,6 @@
 cmake_minimum_required(VERSION 3.5)
 
 idf_build_get_property(__hack_component_targets __COMPONENT_TARGETS)
-#file(GLOB _component_dirs RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/.." ${CMAKE_CURRENT_SOURCE_DIR}/../*)
-#list(APPEND __hack_component_targets ${_component_dirs})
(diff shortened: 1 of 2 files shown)


h2zero:
Awesome, thanks for this! I've been meaning to cleanup that cmake file for a while.

h2zero:
@AronRubin This looks and works fine I wonder if you could apply the same ardruino macro detection to the other checks in the files, i.e NimBLElog.h.

AronRubin:
@h2zero I would absolutely recommend using macros from sdkconfig.h for conditional compile. The ARDUINO_ARCH_ESP32 is an artifact of target_compile_options within arduino-esp32/CMakeLists.txt. Something in the IDF CMake process would have to manually propagate those over or mark them as imported or interface libraries.

h2zero:
Thanks, I will update the rest of the file that use that macro.

PR #61, opened 2021-10-14
Fix some issues on ESP32-S3

huming2207:
Hi @h2zero 

Here's a quick fix to get this library working on ESP32-S3. I suspect there will be more place to fix but so far it (should) just suits my application.

Meanwhile I've also found out that the way you handle data for each characteristic may get in undefined behaviour on ESP32-S3 when the received data is longer than 16 bytes. This is probably caused by Short String Optimisation. But I haven't really looked into it yet. Instead, I used vector instead, as it's more suitable for handling both binary and string data anyway.

To be more specific your master branch, if I type in a few "aaaaa..." and log the buffer via `ESP_LOG_HEXDUMP()` at here: https://github.com/h2zero/esp-nimble-cpp/blob/master/src/NimBLECharacteristic.cpp#L305 and then log the `characteristic->getValue().data()` in the callback class:

It should be always`61 61 61 61 61 61 ...`,   but it shows some garbage data like `14 00 ce 3f 14 00 ce 3f `:

```
W (92228) nimble_cpp: 0x3fca269b   61 61 61 61 61 61 61 61  61 61 61 61 61 61 61 61  |aaaaaaaaaaaaaaaa|
W (92228) cb_class: 0x3fcea328   14 00 ce 3f 14 00 ce 3f  61 61 61 61 61 61 61 61  |...?...?aaaaaaaa|
```

But if the data sent in is short enough, it has the right result:

```
W (36878) nimble_cpp: 0x3fca269b   61 61                                             |aa|
W (36879) cb_class: 0x3fca4848   61 61                                             |aa|
```

So, anyway, like I said before, I decided to scrap the `std::string` and use `vector` instead. I've also called `reserve()` at constructor so that it should somewhat reduce the reallocation and heap fragmentation.

Regards,
Jackson

### DIFF
--- a/src/NimBLECharacteristic.cpp
+++ b/src/NimBLECharacteristic.cpp
@@ -247,7 +247,6 @@ std::vector<uint8_t> NimBLECharacteristic::getValue(time_t *timestamp) {
 
 /**
  * @brief Retrieve the current value of the characteristic, copy to designated buffer to avoid excessive heap allocation
- * @return A std::string containing the current characteristic value.
  */
 void NimBLECharacteristic::getValue(uint8_t *bufOut, size_t *lenOut, size_t bufSize, time_t *timestamp) {
     if (bufOut == nullptr || lenOut == nullptr) return;
--- a/src/NimBLECharacteristic.cpp
+++ b/src/NimBLECharacteristic.cpp
@@ -48,7 +48,7 @@ NimBLECharacteristic::NimBLECharacteristic(const NimBLEUUID &uuid, uint16_t prop
     m_properties  = properties;
     m_pCallbacks  = &defaultCallback;
     m_pService    = pService;
-    m_value       = "";
+    m_value       = std::vector<uint8_t>(BLE_ATT_ATTR_MAX_LEN);
     m_valMux      = portMUX_INITIALIZER_UNLOCKED;
     m_timestamp   = 0;
     m_removed     = 0;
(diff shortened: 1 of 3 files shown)


huming2207:
@h2zero thanks for your reply, this library made my life much easier. 

> Interesting that the hex dump works that way with std::string though, I've not seen this before.

I've also tried using `getValue().c_str()` in the `ESP_LOG_HEXDUMP()` and it gave me the same result.

Meanwhile I think it also relies on too many C++ STL libraries and unnecessary copies. It may cause memory fragmentation as well as getting the code size bloated. I'm also trying to mod it to remove something more, and probably eventually only use `std::vector` but not any other STL library.

h2zero:
If you'd like to see what I'm going for in the future here is a somewhat working PR that I made on the arduino repo. https://github.com/h2zero/NimBLE-Arduino/pull/286

I only made the branch there as I was testing using it for other devices as well that are more RAM limited and it helped immensely. It would be fine here also.

h2zero:
@huming2207 Thank you for your contribution. I have merged your commit for S3 support.

I am going to decline the other commits in favor of #67, If you would like to test that I would appreciate your thoughts.

huming2207:
Thanks for letting me know. Great work on that wrapper class, it's at least much better than `std::string` itself!

Meanwhile, I think it should be possible to get the notification/indication splitting out different data (instead of sharing the same buffer with the "main" characteristic read/write data container). That may be better for some scenarios where I both need to indicate other devices frequently and let other devices read/write my device's characteristics. I was planning to try implementing that during the new year holiday but it was postponed because I need to deal with other personal matters.

h2zero:
Thanks for your comments, hope you'll have time for your projects in the near future.

>I think it should be possible to get the notification/indication splitting out different data (instead of sharing the same buffer with the "main" characteristic read/write data container). That may be better for some scenarios where I both need to indicate other devices frequently and let other devices read/write my device's characteristics.

This capability has been implemented already in #63, though it will need an update with the new class.

PR #63, opened 2021-12-18
Add method to send notifications/indications with custom values

DavidJRobertson:
This is useful when not treating the characteristic as a normal resource and instead using it to send a stream of data.


h2zero:
Looks good to me, thanks!

PR #65, opened 2021-12-31
Update NimBLEServer.h

davidlehrian:
Add #define onMtuChange onMTUChange to support backward compatability with Arduino BLE method name.


h2zero:
Thanks!

PR #67, opened 2022-01-14
Add NimBLEAttValue class.

h2zero:
This is a specialized container class to hold BLE attribute values.

- Removes the use of std::string previously used to store the values.
- Allows for setting/getting/notifying values from std::string, std::vector<uint8_t>, Arduino String, const char*, and uint8_t buffers.
- Has operators retrieve the value as std::string, Arduino String, std::vector<uint8_t>, uint8_t* or char pointers.
- Includes iterators and subscript/random access operator.
- Introduces a max length parameter to the creation of server characteristics/descriptors to limit the size of the memory footprint.
- Nearly Seamless integration with existing code.
- Adds a config option to enable/disable timestamp storage when the value is updated.
- Adds a config option to specify the initial size of the value container if not specified in the constructor.

### Writing POD types to a remote characteristic:
- `pRemoteChar->writeValue(10);`
- `pRemoteChar->writeValue("TEST");`
- `pRemoteChar->writeValue(23.45);`
- `pRemoteChar->writeValue('X');`

### Write container types to a remote characteristic:
- `pRemoteChar->writeValue(String("test string");`
- `pRemoteChar->writeValue(std::string("test std::string");`
- `pRemoteChar->writeValue(std::vector<uint8_t>({1,2,3,4,5,6}));`
- `pRemoteChar->writeValue(NimBLEAttValue({1,2,3,4,5,6}));`

### Reading POD types from a remote characteristic:
- `int i_val =  pRemoteChar->readValue<int>();`
- `double d_val =  pRemoteChar->readValue<double>();
- `const uint8_t* u8p = pRemoteChar->readValue();

### Reading container types from a remote characteristic:
- `std::string ss_str(pRemoteChar->readValue());`
- `String ar_str = pRemoteChar->readValue();`
- `std::vector<uint8_t> vec = pRemoteChar->readValue();`
- `NimBLEAttValue att_val = pRemoteChar->readValue();`

### Reading a single byte from the value:
`uint8_t val =  pRemoteChar->readValue()[2];`

### Printing the hex bytes of a value:
```
for (auto it :  pRemoteChar->readValue()) {
    printf("value: %02x\n", it);
 }
```

Note: The server attributes work the same way but with `setValue()`/`getVaue()`

### Directly using the NimBLEAttValue class:
```
// create an attribute value with a maximum capacity of 8 bytes, initialized to "TEST" (an extra byte will be allocated for the NULL) 
NimBLEAttValue my_att_val("TEST", 8); 

my_att_val += "1234"; // Value now contains "TEST1234"
printf("my_att_val = %s\n", my_att_val.c_str()); // use the c_str() method, print's "TEST1234"

// Concatenate a String and a NimBLEAttValue (same use for std::string)
String test_str = "my att value is: ";
test_str += my_att_val;
Serial.println(test_str);

```

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -49,5 +49,24 @@ config NIMBLE_CPP_ENABLE_ADVERTISMENT_TYPE_TEXT
         Enabling this option will display advertisment types recieved
         while scanning as text messages in the debug log. 
         This will use approximately 250 bytes of flash memory.
+
+config NIMBLE_CPP_ATT_VALUE_TIMESTAMP_ENABLED
+    bool "Enable timestamps to be stored with attribute values."
+    default "n"
+    help
+        Enabling this option will store the timestamp when an attribute value is updated.
+        This allows for checking the last update time using getTimeStamp()
+        or getValue(time_t*). If disabled, the timestamp returned from these functions will be 0.
+        Disabling timestamps will reduce the memory used for each value.
+
+config NIMBLE_CPP_ATT_VALUE_INIT_LENGTH
+    int "Initial attribute value size (bytes) for empty values."
+    range 1 512
+    default 20
+    help
+        Sets the default allocation size (bytes) for each attribute if not specified
+        when the constructor is called. This is also the size used when a remote
+        characteristic or descriptor is constructed before a value is read/notifed.
+        Increasing this will reduce reallocations but increase memory footprint.
         
 endmenu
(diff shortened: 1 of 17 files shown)


h2zero:
Looking for testers on this. Comments/suggestions/bug reports etc.. 

Attn: Template wizards, I'd love to hear your thoughts.

h2zero:
Merging this as I have seen no issues so far in long term testing.

PR #75, opened 2022-04-30
[Breaking] Remove NimBLESecurity class.

h2zero:


h2zero:
Security isn't being removed, just the class 😀.

Security has always been built into the client and server models, with the shared global security settings handled in NimBLEDevice.

This class only existed to be compatible with the original library api to assist users in migration.

I don't feel it's necessary for the security class to still exist anymore here. Anyone that wanted to port their old library code to NimBLE has probably done so already.

andresmanelli:
I kind of saw this coming :smile: Thanks !

PR #97, opened 2022-09-05
Add conn_handle to NimBLECharacteristic::notify to send notification/indication to a specific client

gluhovsky:
@param[in] conn_handle The handle of the connection to a specific subscribed client,
if negative (default) sends notification/indication to the all subscribed clients.

### DIFF
--- a/src/NimBLECharacteristic.cpp
+++ b/src/NimBLECharacteristic.cpp
@@ -1,7 +1,7 @@
 /*
  * NimBLECharacteristic.cpp
  *
- *  Created: on Aug 25, 2022
+ *  Created: on Aug 27, 2022
  *      Author Gluhovsky
  *
  * NimBLECharacteristic.cpp
@@ -26,15 +26,6 @@
 #define NULL_HANDLE (0xffff)
 #define NIMBLE_SUB_NOTIFY   0x0001
 #define NIMBLE_SUB_INDICATE 0x0002
-
-// Quick swap, based on std::move
-template<typename T>
-void swap(T& x, T& y)
-{
-  T z = std::move(x);
-    x = std::move(y);
-    y = std::move(z);
-}
 
 static NimBLECharacteristicCallbacks defaultCallback;
 static const char* LOG_TAG = "NimBLECharacteristic";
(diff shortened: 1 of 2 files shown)


src/NimBLECharacteristic.cpp
```diff
@@ -1,4 +1,9 @@
 /*
+ * NimBLECharacteristic.cpp
+ *
+ *  Created: on Aug 27, 2022
+ *      Author Gluhovsky
+ *
```
h2zero, line 6: Please don't change this like that. You can add a "modified" section with your signature if you like.

src/NimBLECharacteristic.h
```diff
@@ -1,4 +1,9 @@
 /*
+ * NimBLECharacteristic.cpp
+ *
+ *  Created: on Aug 27, 2022
+ *      Author Gluhovsky
+ *
```
h2zero, line 6: Same as above.

src/NimBLECharacteristic.h
```diff
@@ -83,9 +88,9 @@ class NimBLECharacteristic {
     void              indicate();
     void              indicate(const uint8_t* value, size_t length);
     void              indicate(const std::vector<uint8_t>& value);
-    void              notify(bool is_notification = true);
-    void              notify(const uint8_t* value, size_t length, bool is_notification = true);
-    void              notify(const std::vector<uint8_t>& value, bool is_notification = true);
+    void              notify(bool is_notification = true, int16_t conn_handle = (-1));
+    void              notify(const uint8_t* value, size_t length, bool is_notification = true, int16_t conn_handle = (-1));
+    void              notify(const std::vector<uint8_t>& value, bool is_notification = true, int16_t conn_handle = (-1));
```
h2zero, line 93: Instead of `int16_t` it would be better if we used `uint16_t` to match the types. The default value should be the max defined + 1, which is `BLE_HCI_LE_CONN_HANDLE_MAX`.

src/NimBLECharacteristic.cpp
```diff
@@ -459,6 +467,11 @@ void NimBLECharacteristic::notify(const uint8_t* value, size_t length, bool is_n
     int rc = 0;
 
     for (auto &it : m_subscribedVec) {
+        // check if need a specific client
+        if ((conn_handle >= 0) && (it.first != conn_handle)) {
```
h2zero, line 471: We should check for `(conn_handle > BLE_HCI_LE_CONN_HANDLE_MAX)` instead of `(conn_handle >= 0)` here. Please see the comment below about the default value for details.

h2zero (commented):
Thanks, this looks pretty good! Just a couple minor comments.

src/NimBLECharacteristic.h
gluhovsky: corrected as per comment

src/NimBLECharacteristic.cpp
gluhovsky: corrected as per comment

src/NimBLECharacteristic.h
gluhovsky: corrected as per comment

h2zero:
Thanks for making those changes, LGTM!

PR #102, opened 2022-09-13
Remove deprecated API in case of ESP-IDF V5.

armandas:
`esp_nimble_hci_and_controller_deinit` function has been [removed from the API](https://github.com/espressif/esp-idf/blob/0a93ee1337c15668c7f9998a53822f281a17670e/docs/en/migration-guides/release-5.x/bluetooth-low-energy.rst) in ESP IDF version 5.0. Add a compile guard to prevent build errors.

### DIFF
--- a/src/NimBLEDevice.cpp
+++ b/src/NimBLEDevice.cpp
@@ -925,10 +925,12 @@ void NimBLEDevice::deinit(bool clearAll) {
     if (ret == 0) {
         nimble_port_deinit();
 #ifdef ESP_PLATFORM
+#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0)
         ret = esp_nimble_hci_and_controller_deinit();
         if (ret != ESP_OK) {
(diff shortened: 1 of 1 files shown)


h2zero:
Yes, that function is not used, however direct initialization of the controller/host stack is. It's my understanding that this has been reduced to a single call of `nimble_port_init` as the controller/host are now initialized there. 

I'm not sure what effect this has on the existing code, so I'll need to look into it.

armandas:
@h2zero Sorry about the delay. I finally had some time to check this. Indeed, with the original initialization code, I was getting a `E (2391) BLE_INIT: controller init failed` error and continuous rebooting.

Wrapping that with the same check has fixed the issue and it works as expected.

Could you check the updated changes?

walterjj:
Thanyou @armandas !! your fix saved my day!! I works!

h2zero:
Sorry for the delay on this, I've been extremely busy and completely forgot about it 👎. Thanks for this!

PR #109, opened 2023-01-27
//warning: variable 'pkey' set but not used [-Wunused-but-set-variable]

f-hoepfinger-hr-agrartechnik:
//warning: variable 'pkey' set but not used [-Wunused-but-set-variable]


h2zero:
Fixed in #123

PR #111, opened 2023-02-13
Return a pointer to the Value object

tenderlove:
Otherwise we get a copy, and the copy seems to clobber the data we want

Before this commit, calling `getValue()` would return a copy of the underlying `NimBLEAttValue` object.  It seems like the copy doesn't necessarily copy the underlying data stored in the value.  In this commit, I changed it to just return a pointer (which fixes my application), but maybe the class should know how to copy its underlying data.

I'm not sure if this is the best fix (I'm not a pro C++ developer), but as I said this fixes my app.

Thanks!


h2zero:
Providing a pointer to the underlying data structure is dangerous because it could change at any time, thus the reason for the copy. Could you provide an example of the data being "clobbered" to help find the cause?

h2zero:
Closing this as it's not necessary.

PR #117, opened 2023-05-08
Implement a means of asynchronous pin injections for Numeric Comparison and PassKey Input

### LINKED ISSUE #114: Numeric Comparison and Passkey Entry are not properly handled
In `NimBLEServer.cpp`: Currently Numeric Comparison and Passkey Entry (`BLE_SM_IOACT_NUMCMP` and `BLE_SM_IOACT_INPUT` respectively) perform blocking calls to callbacks in order to determine the input for `ble_sm_inject_io`. However, this prevents the SMP Timeout (Vol. 3 Part H Section 3.4 of the Bluetooth 4.2 spec) from properly triggering, as those functions may pend too long waiting for user input.

Furthermore, even if the callbacks do properly timeout themselves, there is no way for them to properly align with the spec, which states that in the event of a timeout "No further SMP commands shall be sent over the L2CAP Security Manager Channel." This is because when the callbacks return, `ble_sm_inject_io` is always called, which results in a "Pairing Failed" command being sent over the SMP channel.

Instead of relying on blocking function calls, these two scenarios should initiate the callbacks only as a means of displaying a prompt to the user, and relying on an asynchronous event for the call to `ble_sm_inject_io`. This allows the timeout event to properly occur.

csmith-morningstar:
Fixes #114

Summary of changes:

- Split `onPassKeyRequest` into two separate functions:
  - `onPassKeyDisplay` which requests the pin (similar to old functionality)
  - `onPassKeyEntry` which notifies the application to prompt the user for a pin number. No return value
- Changed `onConfirmPIN` to lack return value, instead just notifies the app to display the pin number
- Added two functions to the NimBLEServer class: `injectPassKey` and `injectConfirmPIN`, which should be called when the user enters the PIN or presses confirm/cancel respectively.

### DIFF
--- a/examples/Advanced/NimBLE_Server/main/main.cpp
+++ b/examples/Advanced/NimBLE_Server/main/main.cpp
@@ -44,18 +44,26 @@ class ServerCallbacks: public NimBLEServerCallbacks {
     
 /********************* Security handled here **********************
 ****** Note: these are the same return values as defaults ********/
-    uint32_t onPassKeyRequest(){
-        printf("Server Passkey Request\n");
+    uint32_t onPassKeyDisplay(){
+        printf("Server Passkey Display\n");
         /** This should return a random 6 digit number for security 
          *  or make your own static passkey as done here.
          */
         return 123456; 
     };
 
-    bool onConfirmPIN(uint32_t pass_key){
+    void onPassKeyEntry(const BLEAddress& address){
+        printf("Server Passkey Entry\n");
+        /** This should prompt the user to enter the passkey displayed
+         * on the peer device.
+         */
+        NimBLEDevice()::getServer()->injectPassKey(address, 123456)
+    };
+
+    void onConfirmPIN(const BLEAddress& address, uint32_t pass_key){
         printf("The passkey YES/NO number: %d\n", pass_key);
-        /** Return false if passkeys don't match. */
-        return true; 
+        /** Inject false if passkeys don't match. */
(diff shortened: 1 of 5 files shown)


h2zero:
Thanks for this, I will test it shortly 😄

h2zero:
@csmith-morningstar Sorry for taking so long with this. I have thought about it and there are a few things I would like to change.

- The callback signature I believe would be beneficial to have a `NimBLEConnInfo` instance rather than just the address.
- An overload of the `InjectXXX()` functions using a connection handle input would be nice to have.
- The functions could also be moved to `NimBLEDevice` scope so that they could be used by the client code as well and convienience wrappers added to `NimBLEClient` and `NimBLEServer`.

Thanks again for submitting this, it's been a long time coming to fix it 😄

finger563:
I believe this is required to properly implement Google Fast Pair Service - as I've noted in my component example [here](https://github.com/esp-cpp/espp/blob/main/components/gfps_service/example/main/gfps_service_example.cpp). The way I've implemented it with bluedroid is to not respond there, simply store the key, and then respond in the side channel as I've noted [here](https://github.com/esp-cpp/espp/blob/main/components/gfps_service/src/nearby_ble.cpp#L267).

Relatedly, I'm not getting `BLE_GAP_EVENT_PASSKEY_ACTION` in my custom gap event handler (but I am getting the other event), is that expected? Relevant code [here](https://github.com/esp-cpp/espp/blob/main/components/gfps_service/src/nearby_ble.cpp#L82)

PR #134, opened 2023-10-06
Update NimBLEAdvertising.cpp

davidlehrian:
Null terminate the m_name vector because for some reason if you shorten the advertising name it will keep the length the same as the longest name that has been set until the uP is restarted. This means that if you set the name to "longest" and then set it to "short" it will advertise as "shortst" (note two letters from the end of "longest".


h2zero:
@davidlehrian Thanks for this but I believe #166 is a better fix. Please test it and le me know.

davidlehrian:
Setting of the name to a shorter name is working now in my testing. Agreed #166 is a better fix.

PR #143, opened 2024-02-29
Fix inconsistent use of time units

psitto:
Correct times when values in milliseconds are being referred to as if they were in seconds. In examples, documentation and comments.


h2zero:
Thanks for this! Been meaning to get around to some of these fixes.

PR #147, opened 2024-03-13
Change advertising callback from raw pointer to std::function

finger563:
Updates the interface to use a typedef'd std::function for the advertise callback, which is backwards compatible but also allows std::bind and lambda functions.


h2zero:
Thanks for this, this looks like it adds more than just the feature stated, could we split that into a separate PR?

finger563:
Sure - i'll close this and reopen with a simple PR for this specific feature.

PR #156, opened 2024-05-04
feat(NimBLEClient): allow connection id / established flag to be set via public API.

finger563:
This PR replaces #152 

Main use case is to allow an easy mechanism for GATT Servers to read the name of connected clients by passing their connection info to the Client class.

Example (from [espp::BleGattServer](https://github.com/esp-cpp/espp/blob/main/components/ble_gatt_server/include/ble_gatt_server.hpp#L513)):

```c++
  std::string get_connected_device_name(NimBLEConnInfo &conn_info) {
    if (!server_) {
      logger_.error("Server not created");
      return {};
    }
    if (!client_) {
      logger_.error("Client not created");
      return {};
    }

    // since this connection is handled by the server, we won't manually
    // connect, and instead inform the client that we are already connected
    // using this conn handle
    client_->clearConnection(); // make sure any previous connection state is cleared
    client_->setConnection(conn_info);

    // could alternatively use:
    // client_->clearConnection();
    // client_->setConnection(conn_info.getConnHandle(), conn_info.getAddress());

    // refresh the services
    client_->getServices(true);
    // now get Generic Access Service
    auto gas = client_->getService(NimBLEUUID("1800"));
    if (!gas) {
      logger_.error("Failed to get Generic Access Service");
      return {};
    }
    // now get the Device Name characteristic
    auto name_char = gas->getCharacteristic(NimBLEUUID("2A00"));
    if (!name_char) {
      logger_.error("Failed to get Device Name characteristic");
      return {};
    }
    // make sure we can read it
    if (!name_char->canRead()) {
      logger_.error("Failed to read Device Name characteristic");
      return {};
    }
    // and read it
    return name_char->readValue();
  }
```

### DIFF
--- a/src/NimBLEClient.cpp
+++ b/src/NimBLEClient.cpp
@@ -589,7 +589,7 @@ bool NimBLEClient::setConnection(NimBLEConnInfo &connInfo) {
  * @note If the client is already connected to a peer, this will return false.
  * @note This will look up the peer address using the connection id.
  */
-bool NimBLEClient::setConnnection(uint16_t conn_id) {
+bool NimBLEClient::setConnection(uint16_t conn_id) {
     if (isConnected() || m_connEstablished) {
         NIMBLE_LOGE(LOG_TAG, "Already connected");
         return false;
--- a/src/NimBLEClient.cpp
+++ b/src/NimBLEClient.cpp
@@ -564,7 +564,7 @@ uint16_t NimBLEClient::getConnId() {
  *       enables the GATT Server to read the name of the device that has
  *       connected to it.
  */
-bool NimBLEClient::setConnection(const NimBLEConnInfo &connInfo) {
+bool NimBLEClient::setConnection(NimBLEConnInfo &connInfo) {
     if (isConnected() || m_connEstablished) {
         NIMBLE_LOGE(LOG_TAG, "Already connected");
         return false;
--- a/src/NimBLEClient.cpp
+++ b/src/NimBLEClient.cpp
@@ -578,7 +578,7 @@ bool NimBLEClient::setConnection(NimBLEConnInfo &connInfo) {
 } // setConnection
 
 /**
- * @brief Set the connection id for this client.
+ * @brief Set the connection information for this client.
  * @param [in] conn_id The connection id.
  * @note Sets the connection established flag to true.
  * @note This is designed to be used when a connection is made outside of the
@@ -589,7 +589,7 @@ bool NimBLEClient::setConnection(NimBLEConnInfo &connInfo) {
  * @note If the client is already connected to a peer, this will return false.
  * @note This will look up the peer address using the connection id.
  */
-bool NimBLEClient::setConnId(uint16_t conn_id) {
+bool NimBLEClient::setConnnection(uint16_t conn_id) {
     if (isConnected() || m_connEstablished) {
         NIMBLE_LOGE(LOG_TAG, "Already connected");
         return false;
@@ -619,7 +619,7 @@ bool NimBLEClient::setConnId(uint16_t conn_id) {
 } // setConnId
 
 /**
- * @brief Set the connection id for this client.
+ * @brief Set the connection information for this client.
  * @param [in] conn_id The connection id.
  * @param [in] peerAddress The address of the peer that this client is
  * @note Sets the connection established flag to true.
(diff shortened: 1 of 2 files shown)


src/NimBLEClient.cpp
```diff
@@ -552,6 +552,21 @@ uint16_t NimBLEClient::getConnId() {
...
+ * @param [in] conn_id The connection id.
+ * @note Sets the connection established flag to true.
+ * @note This is designed to be used when a connection is made outside of the
+ *       NimBLEClient class, such as when a connection is made by the
+ *       NimBLEServer class and the client is passed the connection id. This use
+ *       enables the GATT Server to read the name of the device that has
+ *       connected to it.
+ */
+void NimBLEClient::setConnId(uint16_t conn_id) {
+    m_conn_id = conn_id;
```
h2zero, line 566: I think we need to add a check here to make sure the client instance is not already connected to a peer and return false if it is. We should also set the peer address here as well which can be retrieved with `ble_gap_conn_find` using the conn_id, or we can add it as another parameter to the function.

h2zero (changes requested):

src/NimBLEClient.cpp
finger563: The check for `isConnected()` would prevent the `Client` object from being reused with a server, since once `setConnId(...)` is called, `isConnected()` would return true. There would be no way to then reset that state. Would you be open to adding a `reset()` function or something similar which clears the `m_conn_id`, `m_connEstablished`, and `m_peerAddress` members for this use case?

finger563: Had the same thought and decided to make that update :)

I did end up renaming it to `setConnection(...)` overloads. Not sure _why_ you'd pass anything other than the `NimBLEConnInfo`, but I think there may be use-cases.

Regarding your question, it was merely to remove the need for multiple allocations / deallocations for what I figured was a pretty simple use-case. That being said, I'm not opposed to going that route.

src/NimBLEClient.cpp
```diff
@@ -552,6 +552,104 @@ uint16_t NimBLEClient::getConnId() {
...
+ * @note This is designed to be used when a connection is made outside of the
+ *       NimBLEClient class, such as when a connection is made by the
+ *       NimBLEServer class and the client is passed the connection id. This use
+ *       enables the GATT Server to read the name of the device that has
+ *       connected to it.
+ * @note If the client is already connected to a peer, this will return false.
+ * @note This will look up the peer address using the connection id.
+ */
+bool NimBLEClient::setConnection(uint16_t conn_id) {
+    if (isConnected() || m_connEstablished) {
```
h2zero, line 593: We can remove this block.

src/NimBLEClient.cpp
```diff
@@ -552,6 +552,104 @@ uint16_t NimBLEClient::getConnId() {
...
+    // we weren't provided the peer address, look it up using ble_gap_conn_find
+    NimBLEConnInfo connInfo;
+    int rc = ble_gap_conn_find(m_conn_id, &connInfo.m_desc);
+    if (rc != 0) {
+        NIMBLE_LOGE(LOG_TAG, "Connection info not found");
+        return false;
+    }
+
+    // ensure the address is valid
+    auto address = connInfo.getAddress();
```
h2zero, line 607: All the code from here down can be removed and replaced with a single call to `setConnection(connInfo);` :)

src/NimBLEClient.cpp
```diff
@@ -552,6 +552,104 @@ uint16_t NimBLEClient::getConnId() {
...
+ * @param [in] peerAddress The address of the peer that this client is
+ * @note Sets the connection established flag to true.
+ * @note This is designed to be used when a connection is made outside of the
+ *       NimBLEClient class, such as when a connection is made by the
+ *       NimBLEServer class and the client is passed the connection id. This use
+ *       enables the GATT Server to read the name of the device that has
+ *       connected to it.
+ * @note If the client is already connected to a peer, this will return false.
+ */
+bool NimBLEClient::setConnection(uint16_t conn_id, const NimBLEAddress &peerAddress) {
```
h2zero, line 633: Let's remove this function as I don't see there being much of a use case for it.

src/NimBLEClient.cpp
h2zero: > Regarding your question, it was merely to remove the need for multiple allocations / deallocations for what I figured was a pretty simple use-case. That being said, I'm not opposed to going that route.

I would say we should create a clearConnection function in this case, couldn't hurt.

src/NimBLEClient.cpp
finger563: So we don't need to check to ensure the address is valid from the looked-up conn_id? I was actually not sure if we should add a similar check within the `setConnection(NimBLEConnInfo &conn_info)` overload, since it does no validity check on the object passed in

h2zero: The address doesn't matter much, as long as the connection ID is valid things should work fine, the address is more for logging etc..

h2zero (commented):
Looks good, thanks! Have you tested this with the changes?

finger563:
> Looks good, thanks! Have you tested this with the changes?

Yup, I've tested my changes locally in espp:

CleanShot 2024-06-04 at 10 41 19

With the output from `espp/components/hid_service/example`:
CleanShot 2024-06-04 at 10 44 18

Showing that the code to get the remote device name using this code is good :)

PR #158, opened 2024-05-04
feat(NimBLEDevice): deleteAllBonds() return code

finger563:
Allows caller to know whether `NimBLEDevice::deleteAllBonds()` (and thus the underlying `ble_store_clear()` call) succeeded or failed.

### DIFF
--- a/src/NimBLEDevice.cpp
+++ b/src/NimBLEDevice.cpp
@@ -567,10 +567,11 @@ int NimBLEDevice::getNumBonds() {
 
 /**
  * @brief Deletes all bonding information.
+ * @returns 0 on success, non-zero on failure.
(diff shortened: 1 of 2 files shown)


src/NimBLEDevice.cpp
```diff
@@ -567,10 +567,11 @@ int NimBLEDevice::getNumBonds() {
 
 /**
  * @brief Deletes all bonding information.
+ * @returns 0 on success, non-zero on failure.
```
h2zero, line 570: To be consistent with the rest of the code we should just return true/false. The error code can be printed in a log message for debugging.

h2zero:
LGTM, thanks!

Issue #159, opened 2024-05-15
Bonded centrals not able to reconnect


cmorganBE:
esp32-s3, esp-idf v5.2, nimble only ble

I have a device that I'd like to run in two modes:

- Previously bonded centrals can connect only
- New devices can pair/connect/bond

This is controlled by a push button with the intent to protect the device from remote tampering by requiring physical interaction with the device to pair/connect/bond new devices.

What I've done so far:
- Switched characteristics over to 'READ | READ_ENC' or 'WRITE | WRITE_ENC'
- NimBLEDevice::setSecurityAuth(true, false true); // bonding = true, mitm = false, secure connection = true
- Enabled Nimble persisting bonding to NVS
- NimBLEAdvertising::setScanFilter(false, false); // at the moment I'm not using whitelists

What I see:
- iphone asks to pair when I open a characteristic on lightblue
- Pairing works, characteristics respond
- Disconnect iphone
- Reconnect iphone
- Connections fail (nothing useful on the esp32 side), lightblue says "failed to connect"

If I want to reconnect I have to go to Settings->Bluetooth and 'Forget' the device.

Using nvs_get I'm seeing the bond in nvs as 'local_irk_1', so that's being persisted.

Is the approach or using setScanFilter() the right one for controlling the connection modes?
Is there something I'm missing on the device side that is preventing reconnections using the bond?

dorianim:
Seems to be the same issue as I have, see:
https://github.com/h2zero/esp-nimble-cpp/issues/11#issuecomment-2116145784

h2zero:
This should be resolved now via #164

PR #164, opened 2024-06-02
Fix failure to reconnect to device when bonded.

### LINKED ISSUE #11: Help with the new security
@h2zero, im converting a project from the old API
which i had security working... 
so the new way to have it going

i have this

NimBLECharacteristic   *MyLEDStatusCharacteristic = pService->createCharacteristic(CHARACTERISTIC_UUID_LED_STATUS, NIMBLE_PROPERTY::READ | NIMBLE_PROPERTY::NOTIFY);
	
which works, to add securtity i would go

NimBLECharacteristic   *MyLEDStatusCharacteristic = pService->createCharacteristic(CHARACTERISTIC_UUID_LED_STATUS, NIMBLE_PROPERTY::READ_ENC | NIMBLE_PROPERTY::NOTIFY);

once i do that, read isnt even showing up as a property for this characteristic

i also had on my 0x2902
MyLEDStatusNotificationDescriptor->setAccessPermissions(ESP_GATT_PERM_READ_ENCRYPTED | ESP_GATT_PERM_WRITE_ENCRYPTED);
so that one cant subscibe with out being authenticated

and this

NimBLESecurity *pSecurity = new NimBLESecurity();
    pSecurity->setStaticPIN(123456);
    pSecurity->setAuthenticationMode(ESP_LE_AUTH_REQ_SC_BOND);

does this simply turn into:

NimBLEDevice::setSecurityPasskey(123456);
    NimBLEDevice::setSecurityAuth(BLE_SM_PAIR_AUTHREQ_BOND);

not getting security to work at all :)

thanks

h2zero:
iOS devices specifically expect to receive the ID key of the bonded device even when RPA is not used. This sets the default value of the key distribution config to provide the ID key as well as the encryption key.

resolves #11 and #159

### DIFF
--- a/src/NimBLEAddress.cpp
+++ b/src/NimBLEAddress.cpp
@@ -139,6 +139,15 @@ uint8_t NimBLEAddress::getType() const {
 
 
 /**
+ * @brief Determine if this address is a Resolvable Private Address.
+ * @return True if the address is a RPA.
+ */
+bool NimBLEAddress::isRpa() const {
+    return (m_addrType && ((m_address[5] & 0xc0) == 0x40));
+} // isRpa
+
+
+/**
  * @brief Convert a BLE address to a string.
  *
  * A string representation of an address is in the format:
--- a/src/NimBLEAddress.h
+++ b/src/NimBLEAddress.h
@@ -43,6 +43,7 @@ public:
     NimBLEAddress(uint8_t address[6], uint8_t type = BLE_ADDR_PUBLIC);
     NimBLEAddress(const std::string &stringAddress, uint8_t type = BLE_ADDR_PUBLIC);
     NimBLEAddress(const uint64_t &address, uint8_t type = BLE_ADDR_PUBLIC);
+    bool            isRpa() const;
     bool            equals(const NimBLEAddress &otherAddress) const;
     const uint8_t*  getNative() const;
     std::string     toString() const;
--- a/src/NimBLEClient.cpp
+++ b/src/NimBLEClient.cpp
@@ -1189,6 +1189,19 @@ int NimBLEClient::handleGapEvent(struct ble_gap_event *event, void *arg) {
             break;
         } //BLE_GAP_EVENT_ENC_CHANGE
 
+        case BLE_GAP_EVENT_IDENTITY_RESOLVED: {
+            NimBLEConnInfo peerInfo;
+            rc = ble_gap_conn_find(event->identity_resolved.conn_handle, &peerInfo.m_desc);
+            if (rc != 0) {
+                NIMBLE_LOGE(LOG_TAG, "Connection info not found");
+                rc = 0;
+                break;
+            }
+
+            pClient->m_pClientCallbacks->onIdentity(peerInfo);
+            break;
+        } // BLE_GAP_EVENT_IDENTITY_RESOLVED
+
         case BLE_GAP_EVENT_MTU: {
             if(pClient->m_conn_id != event->mtu.conn_handle){
                 return 0;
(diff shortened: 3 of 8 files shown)


h2zero:
@cmorganBE Did this work for you?

cmorganBE:
@h2zero I should be testing shortly. I ended up integrating everything and running into wifi init ERR_NO_MEM error even with static stacks and roughly 200k free dram. Moving stacks to spiram and should be able to get everything up and running shortly. For sure I'll let you know how it goes as soon as I can test.

cmorganBE:
@h2zero alright, I'm back looking at this change and had two comments after doing some A-B testing.

Is the example code here that is doing the whitelist add, doing the correct thing?

If I only do:

```
NimBLEDevice::whiteListAdd(connInfo.getAddress());
```

Iphone is not able to reconnect once SetScanFilter(true, true) is called.

I need instead to do:

```
        NimBLEDevice::whiteListAdd(connInfo.getIdAddress());
```

Does this make sense? Since I'm whitelisting the ID and the ID doesn't change or do I need to do both???





In my full app I was seeing iOS prompt to pair each time I read a characteristic. I switched from:

```
NimBLEDevice::setSecurityAuth(false, false, true);
```

to

```
NimBLEDevice::setSecurityAuth(true, false, true);
```

and I'm seeing the device show up again in iOS->Settings->Bluetooth, and not getting prompted to pair each time, BUT switching setSecurityAuth to enable bonding appears to have broken reconnecting when in whitelist mode.

I'm guessing because secure connection is enabled but it seemed unexpected. Thoughts?

h2zero:
Thanks for getting back to me.

So, the addresses are of 2 kinds, random and static. Phones always use a random address that changes every 15 mins or so. When bonding the ID address is exposed to us as well as the IRK. So in your case what "should" work is that after bonding you can add the device ID address to the whitelist, and the scan filter should work, which you have said does. 👍 

Now for :
> BUT switching setSecurityAuth to enable bonding appears to have broken reconnecting when in whitelist mode.

That's not right, maybe you are advertising before the whitelist is loaded somehow?

cmorganBE:
@h2zero let me update or make a new example case in the test repo with bonding enabled and I'll post here when that's done.

NimBLEDevice::whiteListAdd() is inside of onAuthenticationComplete() and the mode switch is:

```
pAdvertising->stop();
pAdvertising->setScanFilter(x, y);
pAdvertising->start();
```

Is there something the user should or has to do related to the whitelists?

h2zero:
That looks reasonable, but there may be some underlying issue.

cmorganBE:
@h2zero alright, so maybe I was premature in indicating it was able to reconnect. I *have* seen it reconnect after bonding when in whitelist mode, however most of the time it doesn't, at least with the test application.

Note all testing has been done with esp-nimble-cpp 'bond-fix' branch.

Here is the updated branch that uses whitelists (note that it doesn't seem to matter what SC setting is used). I also put my test process at the top in the main.cpp file in case there was a mistake in how I'm testing it.

https://github.com/cmorganBE/esp-nimble-cpp-test/tree/bonding_with_whitelist_control_and_sc

h2zero:
Thanks @cmorganBE, I have rebased and pushed a few updates that resolved this as well as added some functionality. Here is the updated main.cpp file that I tested with: 
[main.cpp.txt](https://github.com/user-attachments/files/15807796/main.cpp.txt)

cmorganBE:
@h2zero oooh, this one appears to be working pretty well. Other than an odd unexpected disconnect I was able to connect and reconnect during the whitelist after pairing during the any mode. Looks good!

h2zero:
Great, I'll merge it shortly, thanks!

PR #171, opened 2024-06-25
[NimBLEServer] Add getPeerName and get peer name on connect.

h2zero:
* Adds a new method, getPeerName to NimBLEServer to read the name from the peers device name characteristic.
* Adds a setting to automatically get the name of the peer when connected and provide it as an additional parameter in the onConnect callback


h2zero:
@finger563 You may be interested in this.

finger563:
> @finger563 You may be interested in this.

@h2zero yea, thanks for pointing me to it! I might suggest adding a similar overload to the authentication callback since after auth completes some devices (e.g. iPhones) may change the name they report

h2zero:
Interesting, have you seen this happen before?

finger563:
> Interesting, have you seen this happen before?

Yea - it happens consistently on the first connection / pairing. After connect callback (but before authentication / pairing completes), the iPhone will simply report the name as `iPhone` or `iPhone 14 Plus` for example. After authentication complete, the iPhone will report the actual name of the device, e.g. `William's iPhone`.

h2zero:
Awesome, good to know! I will add that as well!

h2zero:
@finger563 added name to onAuthentication now :)

PR #192, opened 2024-08-07
[BREAKING] Refactor attributes

h2zero:
Refactor attributes to reduce code duplication and improve maintainability.

* Add attribute base classes to provide common code.
* Add const where possible to functions and parameters.
* `NimBLECharacteristic::notify` no longer takes a `bool is_notification` parameter, instead `indicate()` should be called to send indications.
* `NimBLECharacteristic::indicate` now takes the same parameters as `notify`.
* `NimBLECharacteristicCallbacks` and `NimBLEDescriptorCallbacks` methods now take `const NimBLEConnInfo&` instead of non-const.
* `NimBLECharacteristic::onNotify` callback removed as unnecessary, the library does not call notify without app input.
* `NimBLERemoteCharacteristic::getRemoteService` now returns a `const NimBLERemoteService*` instead of non-const.
* Add NimBLEUUID constructor that takes a reference to `ble_uuid_any_t`.
* `NimBLERemoteService::getCharacteristics` now returns a `const std::vector<NimBLERemoteCharacteristic*>&` instead of non-const `std::vector<NimBLERemoteCharacteristic*>*`
* `NimBLERemoteService::getValue` now returns `NimBLEAttValue` instead of `std::string`
* `NimBLEService::getCharacteristics` now returns a `const std::vector<NimBLECharacteristic*>&` instead of a copy of std::vector<NimBLECharacteristic *>.
* Remove const requirement for NimBLEConnInfo parameter in callbacks.
  Const is unnecessary as the data can't be changed by application code.
* Change NimBLERemoteCharacteristic::getRemoteService to return const pointer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,7 @@ idf_component_register(
     "src/NimBLERemoteCharacteristic.cpp"
     "src/NimBLERemoteDescriptor.cpp"
     "src/NimBLERemoteService.cpp"
+    "src/NimBLERemoteValueAttribute.cpp"
     "src/NimBLEScan.cpp"
     "src/NimBLEServer.cpp"
     "src/NimBLEService.cpp"
--- a/docs/Migration_guide.md
+++ b/docs/Migration_guide.md
@@ -383,13 +383,13 @@ The security callback methods are now incorporated in the `NimBLEServerCallbacks
 
 The callback methods are:
 
-> `bool onConfirmPIN(const NimBLEConnInfo& connInfo, uint32_t pin)`
+> `bool onConfirmPIN(NimBLEConnInfo& connInfo, uint32_t pin)`
 
 Receives the pin when using numeric comparison authentication.
 Call `NimBLEDevice::injectConfirmPIN(connInfo, true);` to accept or `NimBLEDevice::injectConfirmPIN(connInfo, false);` to reject.
 <br/>
 
-> `void onPassKeyEntry(const NimBLEConnInfo& connInfo)`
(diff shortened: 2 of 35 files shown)


finger563:
@h2zero I'm not sure why, but after attempting to upgrade to use this code, I'm finding that my GFPS implementation is no longer working. Compilation works just fine (removing the `const` from the auth and passkey callbacks), but for whatever reason the `onConfirmPIN` callback is never being called - and I've found that the event is not even being triggered (the underlying esp-nimble-cpp NimBLEServer::handleGapEvent` is not being called when it should be). I've just started debugging it, but I figured I'd let you know and see if you possibly have some idea of what change might be causing this? The only thing I can think of right now (given that it's the only real change that seems to be coming from this PR) is the change to how attributes are notified/indicated, but I'm not sure...

h2zero:
Thanks for the report, not sure what would have caused that.

finger563:
Did some further testing - it's not just GFPS that has broken - the HID service i have is also broken - it seems to be something about how the `setValue`/`indicate`/`notify` change occurred means that the notifications that i send (calling `notify`) are not being sent to the phone, which manifests as HID inputs not being received by the phone and the GFPS pairing information not being received by the phone either.

finger563:
Confirmed, even just connecting to the example app, and trying to read the `Device Infomation` service's characteristics no longer works (testing using the `BLE Scanner` app on Android).

h2zero:
Thanks, did a quick check myself just now. Looks like the call to send the notification returns with ENOT_CONNECTED, I guess the connection handle is wrong somehow.

PR #194, opened 2024-08-29
fix(NimBLEAdvertising): Ensure `setServiceData()` includes length of data

finger563:
Fork PR here: https://github.com/esp-cpp/esp-nimble-cpp/pull/2


h2zero:
Thanks, looks like there may have been a few misses in the recent refactorings, much appreciated :).

PR #197, opened 2024-09-06
fix(NimBLEDevice): clear all before port_deinit to prevent crash

finger563:
This is an upstream PR matching https://github.com/esp-cpp/esp-nimble-cpp/pull/3. Relevant info copied below:

## Description

Update `NimBLEDevice::deinit()` to run the `clearAll` block (if true) before calling `nimble_port_deinit()`

## Motivation and Context

I found that running `NimBLEDevice::deinit(true)` if I had a client created (as in `espp::BleGattServer`) would lead to a crash as the `~NimBLEClient()` would call `ble_npl_callout_deinit()`, which would lead to a crash. Moving the `clearAll` block to before the `nimble_port_deinit()` call does not result in a crash for the same code.

## How has this been tested?

Building and running in a separate project which was crashing before this change.

See example output here: https://github.com/esp-cpp/espp/pull/325

## Screenshots (if appropriate, e.g. schematic, board, console logs, lab pictures):
Small snippet of the coredump:
CleanShot 2024-09-06 at 08 56 36

### DIFF
--- a/src/NimBLEDevice.cpp
+++ b/src/NimBLEDevice.cpp
@@ -917,6 +917,36 @@ void NimBLEDevice::init(const std::string &deviceName) {
  */
 /* STATIC */
 void NimBLEDevice::deinit(bool clearAll) {
+    if(clearAll) {
+#if defined(CONFIG_BT_NIMBLE_ROLE_PERIPHERAL)
+        if(NimBLEDevice::m_pServer != nullptr) {
+            delete NimBLEDevice::m_pServer;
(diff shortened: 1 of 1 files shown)


h2zero:
Thanks!, LGTM

h2zero:
@finger563 Looks like we have a new issue caused by this change, https://github.com/h2zero/NimBLE-Arduino/issues/724.

I may revert this and look for an alternative, thoughts?

PR #201, opened 2024-10-01
fix: #200 Enable use of `data()`/`size()` before trying `c_str()`/`length()`

### LINKED ISSUE #200: setValue(std::vector<uint8_t>&) fails
`NimBLEAttrValue::setValue(std::vector<uint8_t> s)` fails, likely because it is casting `v` improperly, using `reinterpret_cast<const uint8_t*>(&s)`.

I found this out because my HID code (where I manually create the characteristics) was using this API, though it is disabled on the `NimBLEHIDDevice` class. If you re-enable it (as I have mentioned in the code [here](https://github.com/finger563/nimble-cpp-hid-issue-example/blob/main/main/main.cpp#L91)), then the call to setValue ends up pushing garbage data into the attribute value.

I'd suggest having a code path which checks if the type / container has a `data()` method, similarly to how you check the `c_str()` method, so that you can properly access the data for the container since right now the code will compile with std::vector, but will break.

Note: the link above goes to a simple example repo which has some very basic instructions (comments at the linked part of the code) you should be able to follow to show the failure mode, which is present even on the `NimBLEHIDDevice` implementation.

Note: `data()` is probably a safer / more general check than `c_str()`, so may be better to replace with that anyway.

finger563:
Closes #200 

Replace test for `c_str()` and `length()` (which only matches a few stl containers), for the more general `data()` and `size()` check.

I tested with this example application and it now works properly: 

https://github.com/finger563/nimble-cpp-hid-issue-example/tree/main

Note: I realize I accidentally committed the update to `NimBLEHIDDevice` to expose the `reportMap` characteristic. Personally I think it's a good change, but LMK if you want to revert that.

### DIFF
--- a/src/NimBLEAttValue.h
+++ b/src/NimBLEAttValue.h
@@ -224,25 +224,16 @@ class NimBLEAttValue {
     /**
      * @brief Template to set value to the value of <type\>val.
      * @param [in] s The <type\>value to set.
-     * @param [in] len The length of the value in bytes, defaults to string.length().
+     * @note This function is only availabe if the type T is not a pointer.
      */
     template <typename T>
-    bool setValue(const T& s, uint16_t len = 0) {
+    bool setValue(const T& s) requires (!std::is_pointer_v<T>) {
         if constexpr (Has_data_size<T>::value) {
-            if (len == 0) {
-                len = s.size();
-            }
-            return setValue(reinterpret_cast<const uint8_t*>(s.data()), len);
+            return setValue(reinterpret_cast<const uint8_t*>(s.data()), s.size());
         } else if constexpr (Has_c_str_length<T>::value) {
-            if (len == 0) {
-                len = s.length();
-            }
-            return setValue(reinterpret_cast<const uint8_t*>(s.c_str()), len);
+            return setValue(reinterpret_cast<const uint8_t*>(s.c_str()), s.length());
         } else {
-            if (len == 0) {
-                len = sizeof(s);
-            }
-            return setValue(reinterpret_cast<const uint8_t*>(&s[0]), len);
+            return setValue(reinterpret_cast<const uint8_t*>(&s[0]), sizeof(s));
         }
     }
 
--- a/src/NimBLEAttValue.h
+++ b/src/NimBLEAttValue.h
@@ -234,7 +234,7 @@ class NimBLEAttValue {
         } else if constexpr (Has_c_str_length<T>::value) {
             return setValue(reinterpret_cast<const uint8_t*>(s.c_str()), s.length());
         } else {
-            return setValue(reinterpret_cast<const uint8_t*>(&s[0]), sizeof(s));
+            return setValue(reinterpret_cast<const uint8_t*>(&s), sizeof(s));
         }
     }
 
--- a/src/NimBLEAttValue.h
+++ b/src/NimBLEAttValue.h
@@ -227,7 +227,7 @@ class NimBLEAttValue {
      * @note This function is only availabe if the type T is not a pointer.
      */
     template <typename T>
-    std::enable_if_t<!std::is_pointer_v<T>, bool>
+    std::enable_if<!std::is_pointer_v<T>, bool>::type
     setValue(const T& s) {
         if constexpr (Has_data_size<T>::value) {
             return setValue(reinterpret_cast<const uint8_t*>(s.data()), s.size());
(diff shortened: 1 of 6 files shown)


finger563:
Follow-up: if you're willing to require `c++20`, then we can use the `concepts` and `requires` functionality to clean up the metaprogramming.

h2zero:
Thanks, the problem with this is that it will not work with Arduino String, which is why I used the c_str and length check since those functions are available in std::string and String.

finger563:
@h2zero I've pushed an update which keeps support for `data/size` but will fall back to `c_str/length` and finally `&v[0]/sizeof(v)`

finger563:
FYI, esp-idf 4.4 reached end of life back in July, so it is no longer supported. Might be worth upgrading your CI to allow the codebase to migrate to more modern standards :)

h2zero:
Looks good to me, thanks!. I'll do some testing then merge shortly.

src/NimBLEAttValue.h
```diff
@@ -216,39 +224,18 @@ class NimBLEAttValue {
...
-        if (len == 0) {
-            len = s.length();
+    typename std::enable_if<!std::is_pointer<T>::value, bool>::type
+    setValue(const T& s) {
+        if constexpr (Has_data_size<T>::value) {
+            return setValue(reinterpret_cast<const uint8_t*>(s.data()), s.size());
+        } else if constexpr (Has_c_str_length<T>::value) {
+            return setValue(reinterpret_cast<const uint8_t*>(s.c_str()), s.length());
+        } else {
+            return setValue(reinterpret_cast<const uint8_t*>(&s[0]), sizeof(s));
```
h2zero, line 237: The `[0]` in cast needs to be removed, compilation fails in some tests.

src/NimBLECharacteristic.h
```diff
@@ -77,36 +75,47 @@ class NimBLECharacteristic : public NimBLELocalValueAttribute {
...
-    notify(const T& value, bool is_notification = true) const {
-        notify(reinterpret_cast<const uint8_t*>(value.c_str()), value.length(), is_notification);
+    typename std::enable_if<!std::is_pointer<T>::value, void>::type
+    notify(const T& value, uint16_t conn_handle = BLE_HS_CONN_HANDLE_NONE) const {
+        if constexpr (Has_data_size<T>::value) {
+            notify(reinterpret_cast<const uint8_t*>(value.data()), value.size(), conn_handle);
+        } else if constexpr (Has_c_str_length<T>::value) {
+            notify(reinterpret_cast<const uint8_t*>(value.c_str()), value.length(), conn_handle);
+        } else {
+            notify(reinterpret_cast<const uint8_t*>(&value[0]), sizeof(value), conn_handle);
```
h2zero, line 95: Same as above

src/NimBLECharacteristic.h
```diff
@@ -77,36 +75,47 @@ class NimBLECharacteristic : public NimBLELocalValueAttribute {
...
-    indicate(const T& value) const {
-        indicate(reinterpret_cast<const uint8_t*>(value.c_str()), value.length());
+    typename std::enable_if<!std::is_pointer<T>::value, void>::type
+    indicate(const T& value, uint16_t conn_handle = BLE_HS_CONN_HANDLE_NONE) const {
+        if constexpr (Has_data_size<T>::value) {
+            indicate(reinterpret_cast<const uint8_t*>(value.data()), value.size(), conn_handle);
+        } else if constexpr (Has_c_str_length<T>::value) {
+            indicate(reinterpret_cast<const uint8_t*>(value.c_str()), value.length(), conn_handle);
+        } else {
+            indicate(reinterpret_cast<const uint8_t*>(&value[0]), sizeof(value), conn_handle);
```
h2zero, line 117: here too

src/NimBLERemoteValueAttribute.h
```diff
@@ -88,32 +78,18 @@ class NimBLERemoteValueAttribute : public NimBLEAttribute {
...
-    typename std::enable_if<Has_c_str_len<T>::value, bool>::type
-# endif
-    writeValue(const T& s, bool response = false) const {
-        return writeValue(reinterpret_cast<const uint8_t*>(s.c_str()), s.length(), response);
+        if constexpr (Has_data_size<T>::value) {
+            return writeValue(reinterpret_cast<const uint8_t*>(v.data()), v.size(), response);
+        } else if constexpr (Has_c_str_length<T>::value) {
+            return writeValue(reinterpret_cast<const uint8_t*>(v.c_str()), v.length(), response);
+        } else {
+            return writeValue(reinterpret_cast<const uint8_t*>(&v[0]), sizeof(v), response);
```
h2zero, line 91: and here

finger563:
@h2zero Thanks, I've addressed your comments, hopefully they all pass now :)

h2zero:
Thanks!

PR #204, opened 2024-10-05
Fix check for arduino component

### LINKED ISSUE #202: Include error on new project
Created a minimal project using esp-idf using Arduino as an IDF component following [these instructions](https://docs.espressif.com/projects/arduino-esp32/en/latest/esp-idf_component.html). Using arduino-esp32 3.0.5 and IDF 5.1.4. Then added NimBLE as a component to the project. This results in the following error:

```
fatal error: esp32-hal-bt.h: No such file or directory
```

I was able to make it compile by adding `arduino-esp32` to `ESP_NIMBLE_PRIV_REQUIRES` in the nimble CMakeLists.txt. I am not using platformio.

jefflongo:
Fixes #202. Tested with IDF 5.1.4 / Arduino core 3.0.x and IDF 4.4.8 / Arduino core 2.0.x.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,10 +14,17 @@ elseif("nimble" IN_LIST BUILD_COMPONENTS OR "__nimble" IN_LIST __hack_component_
     )
 endif()
 
+# Arduino install using IDF component manager
 if("espressif__arduino-esp32" IN_LIST BUILD_COMPONENTS OR "___idf_espressif__arduino-esp32" IN_LIST __hack_component_targets)
     list(APPEND ESP_NIMBLE_PRIV_REQUIRES
         arduino-esp32
     )
+# Manual installation of Arduino framework
+elseif("arduino" IN_LIST BUILD_COMPONENTS OR "__idf_arduino" IN_LIST __hack_component_targets)
+    list(APPEND ESP_NIMBLE_PRIV_REQUIRES
+        arduino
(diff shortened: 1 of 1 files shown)


jefflongo:
As a side note, is `__hack_component_targets` really necessary? It doesn't seem necessary for Arduino, but I didn't remove it because I don't know for what cases it was added in the first place.

jefflongo:
Also, the README states that defining `ARDUINO_ARCH_ESP32` is required. Looking at my `compile_commands.json`, it seems to be defined without me defining it in the root CMakeLists.txt. Do you have any evidence that this is still required?

h2zero:
Thanks for this, looks like the issue I had with this was a different problem. That being said, the reason the arduino dependency wasn't being added was simply due to the folder name. If you follow the instructions to use arduino as a component it clones the repo into the "arduino" folder, not "arduino-esp32", which is why this cmake file checks for that.

So, instead of replacing the check, we should check for both cases. 

The  `__hack_component_targets` is a bit of a legacy thing from years ago, probably not needed now but I'll review that soon.

You can ignore that bit in the readme, needs updating too.

jefflongo:
The reason I missed that was because I followed the "Installing using IDF Component Manager" instruction in the Espressif documentation instead of the "Manual installation of Arduino framework" instruction. So you're right, we should be checking for both.

PR #208, opened 2024-10-21
Add initial esp32-P4 support

chegewara:
This commit adds initial support for esp32-p4 with esp32-c6 as a radio.
Configuration was tested on this dev board
https://docs.espressif.com/projects/esp-dev-kits/en/latest/esp32p4/esp32-p4-function-ev-board/user_guide.html

Since it is using esp_hosted component and we cant use bluedroid API i had to disable get and set ble power API.
Maybe someone can find better solution, or maybe in the future we can re-enable those functions, when we have support for bluedroid.

Library tested with ble + wifi configuration and works fine.

PS i added some options to Kconfig to enable bluetooth by default, but there is no option to override espressif configs and to enable nimble by default, so it is last option which we have to do manually or with defconfig in our app

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -76,4 +76,51 @@ config NIMBLE_CPP_DEBUG_ASSERT_ENABLED
         Enabling this option will add debug asserts to the NimBLE CPP library.
         This will use approximately 1kB of flash memory.
 
+#
+# BT config
+#
+config BT_ENABLED
+    bool "Bluetooth"
+    default "y"
+    help
+        Select this option to enable Bluetooth and show the submenu with Bluetooth configuration choices.
+
+
+config BT_NIMBLE_ENABLED
+    bool "NimBLE - BLE only"
+    default "y"
+    help
+        This option is recommended for BLE only usecases to save on memory
+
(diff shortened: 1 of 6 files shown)


h2zero:
Thanks! I think we can use the alternative power get/set functions through NimBLE API that I am working on in #198

Kconfig
h2zero, line 1: Can we make these available only if the P4 is selected?

chegewara: Why?
BT_ENABLE is required for all targets, not only P4, so we have to use menuconfig to make the changes or to use sdkconfig.defaults.

h2zero: I was thinking it may conflict when using other chips where those options are already available, cool if that's not the case though.

chegewara: 1. Options are "overlapped" and when you change in one place in menuconfig, the other one is updated, so you can have as many copies as you want and all will have the same value
2. i tested changes with the same code on P4 and S3, so it should be fine

h2zero: Awesome, thank you 👍

PR #210, opened 2024-10-23
[Server] Implement connection parameters update callback

### LINKED ISSUE #209: Callback for connection params update
When calling `server->updateConnParams`, a request is made which the peer device may agree on, or set to something else. Is there any way to know when the connection parameters have been updated so I can check what they are?

jefflongo:
Closes #209


h2zero:
LGTM Thanks!

PR #225, opened 2024-11-14
Fix NimBLEUtils::dataToHexString returned string, must resize str.

thekurtovic:
Commit 84f4d4f897963559dbb92ca73d1b68402b914c94 has broken NimBLEUtils::dataToHexString(). Manufacturing data is no longer visible when printing NimBLEAdvertisedDevice::toString().
Before
`I [  7210][main.cpp:198] BLE adv (-81 db, phy 1 0) Name: SHIELD, Address: 53:6f:93:26:6e:e7, manufacturer data: , serviceUUID: 0x3456, txPower: -15`

After
`I [  7238][main.cpp:198] BLE adv (-81 db, phy 1 0) Name: SHIELD, Address: 53:6f:93:26:6e:e7, manufacturer data: ffff65636635323533326465643734323730, serviceUUID: 0x3456, txPower: -15`


h2zero:
Thanks! Weird that it works in local testing but not on esp32.

PR #227, opened 2024-11-15
Refactor client connection establishment and client deletion.

h2zero:
* Removes the connection established flag as it should not be necessary.
* Deleting the client instance from the `onDisconnect` callback is now supported.
* `NimBLEDevice::deleteClient` no longer blocks tasks
* Adds a new `Config` struct to `NimBLEClient` to efficiently set single bit config settings.
* Adds `NimBLEClient::setSelfDelete` that takes the bool parameters `deleteOnDisconnect` and `deleteOnConnectFail`
  This will configure the client to delete itself when disconnected or the connection attempt fails.
* Adds `NimBLEClient::setConfig` and `NimBLEClient::getConfig` which takes or returns a `NimBLEClient::Config` object respectively.

* Reword `BLE_HS_EAPP` error string to be more accurate.

### DIFF
--- a/src/NimBLEClient.cpp
+++ b/src/NimBLEClient.cpp
@@ -63,10 +63,7 @@ NimBLEClient::NimBLEClient(const NimBLEAddress& peerAddress)
       m_pClientCallbacks{&defaultCallbacks},
       m_connHandle{BLE_HS_CONN_HANDLE_NONE},
       m_terminateFailCount{0},
-      m_deleteCallbacks{false},
-      m_connEstablished{false},
-      m_asyncConnect{false},
-      m_exchangeMTU{true},
+      m_config{},
 # if CONFIG_BT_NIMBLE_EXT_ADV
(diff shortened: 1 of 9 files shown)


src/NimBLEClient.cpp
```diff
@@ -929,41 +945,45 @@ int NimBLEClient::handleGapEvent(struct ble_gap_event* event, void* arg) {
                     break;
             }
 
+            NIMBLE_LOGD(LOG_TAG, "disconnect; reason=%d, %s", rc, NimBLEUtils::returnCodeToString(rc));
+
+            pClient->m_connHandle         = BLE_HS_CONN_HANDLE_NONE;
```
thekurtovic, line 950: This should be moved to be after the onDisconnect() callback.

h2zero: I think so too, still a work in progress.

PR #231, opened 2024-11-18
[BREAKING] - Refactor NimBLEScan

### LINKED ISSUE #73: NimBLEScan::clearResults doesn't actually release memory space
Hello, I found that when there are too many Bluetooth devices around, the memory usage continues to rise after scanning is turned on.

I called "clearResults" to clean up, but it didn't work,
Then I read the code and found that "clear" is only called for "vector" in "clearResults", which doesn't actually free the memory occupied by the vector.

This is a "vector" based memory management problem.

https://github.com/h2zero/esp-nimble-cpp/blob/a36655c105444f26d1ef7dd0893b189ab26f23ce/src/NimBLEScan.cpp#L507-L513
Add a line of code below "clear" to solve:
`std::vector<NimBLEAdvertisedDevice *>().swap(m_scanResults.m_advertisedDevicesVector);`
```cpp
 void NimBLEScan::clearResults() { 
     for(auto &it: m_scanResults.m_advertisedDevicesVector) { 
         delete it; 
     } 
     m_scanResults.m_advertisedDevicesVector.clear(); 
     std::vector<NimBLEAdvertisedDevice *>().swap(m_scanResults.m_advertisedDevicesVector);
     clearDuplicateCache(); 
 } 
```

h2zero:
* General code cleanup
* `NimBLEScan::start` will no longer clear cache or results if scanning is already in progress.
* `NimBLEScan::clearResults` will now reset the vector capacity to 0.
* `NimBLEScan::stop` will no longer call the `onScanEnd` callback as the caller should know its been stopped when this is called.
* `NimBLEScan::clearDuplicateCache` has been removed as it was problematic and only for the esp32. Stop and start the scanner for the same effect.
* `NimBLEScan::start` takes a new bool parameter `restart`, default `true`, that will restart an already in progress scan and clear the duplicate filter so all devices will be discovered again.
* Scan response data that is received without advertisement first will now create the device and send a callback.
* Added new method: `NimBLEAdvertisedDevice::isScannable()` that returns true if the device is scannable.
* Added default callbacks for `NimBLEScanCallbacks`
* `NimBLEScanCallbacks` function signatures updated:
* - `onDiscovered` now takes a `const NimBLEAdvertisedDevice*`
* - `onResult` now takes a `const NimBLEAdvertisedDevice*`
* - `onScanEnd` now takes a `const NimBLEScanResults&` and `int reason`
* Added new erase overload: `NimBLEScan::erase(const NimBLEAdvertisedDevice* device)`
* `NimBLEScanResults::getDevice` methods now return `const NimBLEAdvertisedDevice*`
* `NimBLEScanResults` iterators are now `const_iterator`

Fixes #73


thekurtovic:
> NimBLEScan::stop will no longer call the onScanEnd callback as the caller should know its been stopped when this is called.

There are some spots where `NimBLEScan::stop` is called internally, I think it would be good to still call the `onScanEnd` callback in such cases. What about adding a parameter which defaults to false, else will call `onScanEnd`?
I don't stop the scan explicitly before connecting as I liked how the `NimBLE Async Client` handles it.

h2zero:
The reason for the change is so the application can make use of the callback to attempt to restart it. When connecting it will always be stopped first and that's about the only time that it is internally called other than when the app specifies to reset.

h2zero:
Hmm, now that I think about it, a reason code parameter should be added to the callback so the application can decide what to do from there.

thekurtovic:
I don't understand how can the application use the callback if it's never called? Do you mean the proper usage would be to always have the application stop the scan explicitly, and not use the convenience of the connect function?
```
scan state 0
Advertised Device found: Name: , Address: e3:e4:5b:75:ad:36
Found Our Device
Connected to: e3:e4:5b:75:ad:36
scan state 1
peer address: e3:e4:5b:75:ad:36
Services:

e3:e4:5b:75:ad:36 Disconnected, reason = 534
scan state 1
scan state 1
```

```
#include <NimBLEDevice.h>

static constexpr uint32_t scanTimeMs = 5 * 1000;
bool scanning = true;

class ClientCallbacks : public NimBLEClientCallbacks {
    void onConnect(NimBLEClient* pClient) {
        printf("Connected to: %s\n", pClient->getPeerAddress().toString().c_str());
    }

    void onDisconnect(NimBLEClient* pClient, int reason) {
        printf("%s Disconnected, reason = %d\n", pClient->getPeerAddress().toString().c_str(), reason);
    }
} clientCBB;

class scanCallbacks : public NimBLEScanCallbacks {
    void onResult(NimBLEAdvertisedDevice* advertisedDevice) {
        printf("Advertised Device found: %s\n", advertisedDevice->toString().c_str());
        if (!advertisedDevice->haveName() && !advertisedDevice->haveServiceData()) {
            printf("Found Our Device\n");

            auto pClient = NimBLEDevice::getDisconnectedClient();
            if (!pClient) {
                pClient = NimBLEDevice::createClient(advertisedDevice->getAddress());
                if (!pClient) {
                    printf("Failed to create client\n");
                    return;
                }
            }

            pClient->setClientCallbacks(&clientCBB, false);
            if (!pClient->connect(true, true, false)) { // delete attributes, async connect, no MTU exchange
                NimBLEDevice::deleteClient(pClient);
                printf("Failed to connect\n");
                return;
            }
        }
    }

    void onScanEnd(NimBLEScanResults results) {
        printf("Scan Ended\n");
        scanning = false;
        NimBLEDevice::getScan()->start(scanTimeMs);
    }
};

void setup()
{
    Serial.begin(115200);
    NimBLEDevice::init("");
    NimBLEDevice::setPower(3); /** +3db */

    printf("scan state %u\n", scanning);

    NimBLEScan* pScan = NimBLEDevice::getScan();
    pScan->setScanCallbacks(new scanCallbacks());
    pScan->setInterval(45);
    pScan->setWindow(15);
    pScan->setActiveScan(true);
    pScan->start(scanTimeMs);
    scanning = true;
}

void loop()
{
    vTaskDelay(1000);
    if (scanning)
        printf("scan state %u\n", scanning);

    auto pClients = NimBLEDevice::getConnectedClients();
    if (!pClients.size()) {
        return;
    }

    for (auto& pClient : pClients) {
        printf("%s\n", pClient->toString().c_str());
        NimBLEDevice::deleteClient(pClient);
    }
}
```

h2zero:
The purpose of the callback is to let the application know when the scan stopped due to a timeout or error. The stack itself doesn't generate an event when the scan is stopped manually, likely for the same reason.

PR #236, opened 2024-11-19
Fix compilation errors in some situations.

### LINKED ISSUE #234: Can't compile on IDF 5.3
Wanted to start testing on 5.3 but I get errors compiling unless I go back to 68b82f5b855040b320c7865153e6289904ef132f.
Build output is really long, I can't make any sense of it.
[output.txt](https://github.com/user-attachments/files/17819081/output.txt)

Testing with this example on esp32-s3.
https://github.com/pioarduino/platform-espressif32/tree/develop/examples/espidf-arduino-h2zero-BLE_scan

5.3 isn't part of the workflow checks so I'm not sure if it's supported yet.

h2zero:
Fixes #234

### DIFF
--- a/src/HIDTypes.h
+++ b/src/HIDTypes.h
@@ -25,9 +25,9 @@
 #define HID_VERSION_1_11    (0x0111)
 
 /* HID Class */
-#define HID_CLASS           (3)
-#define HID_SUBCLASS_NONE   (0)
-#define HID_PROTOCOL_NONE   (0)
+#define BLE_HID_CLASS           (3)
+#define BLE_HID_SUBCLASS_NONE   (0)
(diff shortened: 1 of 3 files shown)


iranl:
I believe this issue was introduced in a recent update of ESP-IDF 5.3. 
There is an issue about this here https://github.com/espressif/esp-idf/issues/14840, which is already fixed by https://github.com/espressif/esp-nimble/commit/ca6c6173fca923cfa1966891fc6623fd9f9cb323.

So not sure if this change on the side of esp-nimble-cpp is needed (at least not long term).

h2zero:
Thanks, it will be needed as this commonly breaks due to NimBLE defining min/max in headers and conflicting anywhere we `#include <string>`. I'm inclined to keep this change here.

Issue #232, opened 2024-11-18
NimBLEClient::getRssi() returning 0 on async connection


thekurtovic:
Printing `NimBLEClient::getRssi` within `onConnect ` for async connection tends to always show 0*. No issue getting RSSI value after non-async connection is made. Oddly there's no error given as `ble_gap_conn_rssi` is returning 0.
Testing with https://github.com/h2zero/esp-nimble-cpp/pull/214

*Occasionally will get a non-zero value but it's rare.

Async
```
D (8884) NimBLEScan: << stop()
D (8884) NimBLEClient: >> connect(e3:e4:5b:75:ad:36)
D (9846) NimBLEClient: Got Client event
I (9846) NimBLEClient: Connected event
I [  9847][main.cpp:144] Connected to e3e45b75ad36, RSSI 0 (bond 1)
D (9862) NimBLEClient: Got Client event
I (9863) NimBLEClient: mtu update: mtu=247
D (9873) NimBLEClientCallbacks: onMTUChange: default
D (9873) NimBLEClient: exchangeMTUCb: status=0, mtu=247
D (9943) NimBLEClient: Got Client event
I (9944) NimBLEClient: Connection parameters updated.
D (9952) NimBLEClient: Got Client event
D (9975) NimBLEClient: Got Client event
D (10297) NimBLEClient: Got Client event
D [ 10297][main.cpp:182] Auth event 1 1 1 16
```

Non-async
```
D (13520) NimBLEScan: << stop()
D (13520) NimBLEScan: erase device: e3:e4:5b:75:ad:36
I [ 13511][main.cpp:354] Create client, attempting to connect...
D (13531) NimBLEClient: >> connect(e3:e4:5b:75:ad:36)
D (15487) NimBLEClient: Got Client event
I (15487) NimBLEClient: Connected event
D (15502) NimBLEClient: Got Client event
I (15503) NimBLEClient: mtu update: mtu=247
D (15503) NimBLEClientCallbacks: onMTUChange: default
D (15513) NimBLEClient: exchangeMTUCb: status=0, mtu=247
D (15513) NimBLEClient: Got Client event
I (15515) NimBLEClient: Connection established
I [ 15524][main.cpp:144] Connected to e3e45b75ad36, RSSI -40 (bond 1)
D (15540) NimBLEClient: Got Client event
D (15540) NimBLEClient: << connect()
I [ 15552][main.cpp:371] Securing connection...
[...]

PR #241, opened 2024-11-25
Add NimBLEClientCallbacks::onConnectFail callback

h2zero:
Adds a callback that is called when the connection attempt fail while connecting asynchronously.

### DIFF
--- a/src/NimBLEClient.cpp
+++ b/src/NimBLEClient.cpp
@@ -977,28 +977,28 @@ int NimBLEClient::handleGapEvent(struct ble_gap_event* event, void* arg) {
             rc = event->connect.status;
             if (rc == 0) {
                 pClient->m_connHandle = event->connect.conn_handle;
+
+                if (pClient->m_config.asyncConnect) {
+                    pClient->m_pClientCallbacks->onConnect(pClient);
+                }
+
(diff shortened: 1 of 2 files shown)


h2zero:
@thekurtovic Please let me know if this works for you.

src/NimBLEClient.cpp
```diff
@@ -989,15 +989,16 @@ int NimBLEClient::handleGapEvent(struct ble_gap_event* event, void* arg) {
...
+                if (pClient->m_config.deleteOnConnectFail) {
                     delete pClient;
-                    return 0;
                 }
+                return 0;
             }
 
-            if (pClient->m_config.asyncConnect) {
-                pClient->m_pClientCallbacks->onConnect(pClient);
-            } else if (!pClient->m_config.exchangeMTU) {
```
thekurtovic, line 1000: This call to `onConnect` should be left as is for async connects.

h2zero: Oops, I actually meant to move that instead of delete, thanks!

thekurtovic:
LGTM. Tested with esp32-s3 and it resolves recent issues mentioned in https://github.com/h2zero/esp-nimble-cpp/issues/232.

PR #251, opened 2024-12-02
Add support for creating a NimBLEClient from a NimBLEServer peer.

h2zero:
This allows the NimBLEServer instance to create a NimBLEClient instance to read/write form/to a connected peer. Only one instance is supported subsequent calls will overwrite the previous client connection information and data.

### DIFF
--- a/src/NimBLEClient.cpp
+++ b/src/NimBLEClient.cpp
@@ -537,62 +537,6 @@ uint16_t NimBLEClient::getConnHandle() const {
 } // getConnHandle
 
 /**
- * @brief Clear the connection information for this client.
- * @note This is designed to be used to reset the connection information after
- * calling setConnection(), and should not be used to disconnect from a peer.
- * To disconnect from a peer, use disconnect().
- */
-void NimBLEClient::clearConnection() {
-    m_connHandle  = BLE_HS_CONN_HANDLE_NONE;
-    m_peerAddress = NimBLEAddress{};
-} // clearConnection
-
-/**
- * @brief Set the connection information for this client.
- * @param [in] connInfo The connection information.
- * @return True on success.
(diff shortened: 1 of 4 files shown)


h2zero:
@finger563 You may be interested in reviewing this, your feedback would be greatly appreciated.

h2zero:
That shouldn't be necessary now because the client is created and managed by the server and `NimBLEDevice::deint(true)` will only manage the clients created through that interface. The server will delete the client instance when it is deleted so there should be no issues.

h2zero:
@finger563 Did you perhaps test this at all?

finger563:
@h2zero Unfortunately, I've not had time to yet. If you need, I'll be able to test late today or tomorrow.

finger563:
@h2zero one change in behavior I'm seeing when using the new client instead of the one I was using before is that the initial connection (before auth is complete) now returns failure to get generic access service from the client:

Disconnect and reconnect and it works fine.

CleanShot 2024-12-05 at 08 07 37

So that's something that ideally could be fixed?

h2zero:
Thanks, not sure why that errored, would need to see logging with info messages on.

finger563:
@h2zero 

Here are some additional logs turned on esp-nimble-cpp:
CleanShot 2024-12-05 at 10 13 59

h2zero:
Thanks, found the issue.

finger563:
@h2zero Thanks, confirmed it's fixed :)
CleanShot 2024-12-05 at 13 19 01

finger563 (approved):
Works well in my testing :)

h2zero:
Great, thanks!  I'm considering removing the server name read to go along with this as it seems redundant and bloats the code.

PR #252, opened 2024-12-03
Add a workaround for esp32s3 and esp32c3 tx power bug

h2zero:
This adds a workaround to get the tx power when the function returns error due to a bug introduced in some versions of esp-idf.

* Added error checking, return value will be 0xFF if there was an error.

### DIFF
--- a/src/NimBLEDevice.cpp
+++ b/src/NimBLEDevice.cpp
@@ -425,10 +425,19 @@ std::vector<NimBLEClient*> NimBLEDevice::getConnectedClients() {
 #  ifndef CONFIG_IDF_TARGET_ESP32P4
 /**
  * @brief Get the transmission power.
- * @return The power level currently used in esp_power_level_t.
+ * @return The power level currently used in esp_power_level_t or a negative value on error.
  */
 esp_power_level_t NimBLEDevice::getPowerLevel(esp_ble_power_type_t powerType) {
-    return esp_ble_tx_power_get(powerType);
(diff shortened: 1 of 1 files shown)


h2zero:
@thekurtovic Any thoughts on this?

h2zero:
I don't see this occurring outside of the bug since we are always calling the function with a valid parameter.

src/NimBLEDevice.cpp
```diff
@@ -479,15 +479,27 @@ bool NimBLEDevice::setPower(int8_t dbm) {
...
 # ifdef ESP_PLATFORM
 #  ifdef CONFIG_IDF_TARGET_ESP32P4
     return 0xFF; // CONFIG_IDF_TARGET_ESP32P4 does not support esp_ble_tx_power_get
 #  else
     int pwr = esp_ble_tx_power_get(ESP_BLE_PWR_TYPE_DEFAULT);
+
+#   if defined(CONFIG_IDF_TARGET_ESP32C3) || defined(CONFIG_IDF_TARGET_ESP32S3)
+    // workaround for bug when "enhanced tx power" was added
+    if (pwr == 0xFF) {
+        pwr = esp_ble_tx_power_get(ESP_BLE_PWR_TYPE_CONN_HDL3);
```
thekurtovic, line 494: Should this be moved into `getPowerLevel`, and have this function call it?

h2zero: Hmm, yes it probably should 👍

h2zero: done

PR #256, opened 2024-12-05
NimBLELog support logging without IDF logging, allow custom colors.

thekurtovic:
* This will allow NimBLE logging to be enabled without requiring IDF logging to be enabled.
* Additionally this will allow for color codes to be customized for each log level.
* Cleaned up preprocessor indentation style.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -26,6 +26,113 @@ config NIMBLE_CPP_LOG_LEVEL
     default 3 if NIMBLE_CPP_LOG_LEVEL_INFO
     default 4 if NIMBLE_CPP_LOG_LEVEL_DEBUG
 
+config NIMBLE_LOG_OVERRIDE
+    bool "Enable log override, decoupled from general IDF logging."
+    default "n"
+    help
+        Enabling this option will allow NimBLE logging to be printed without
+        being tied to IDF log level.
+
+menu "NIMBLE Log Override Colors"
+    depends on NIMBLE_LOG_OVERRIDE
+
+    choice NIMBLE_LOG_OVERRIDE_COLOR_ERR
+        prompt "NimBLE CPP log override color Error"
+        default NIMBLE_LOG_OVERRIDE_COLOR_ERR_NONE
+        help
+            Select NimBLE CPP log override error color.
+
+        config NIMBLE_LOG_OVERRIDE_COLOR_ERR_NONE
+            bool "None"
+        config NIMBLE_LOG_OVERRIDE_COLOR_ERR_BLACK
+            bool "Black"
+        config NIMBLE_LOG_OVERRIDE_COLOR_ERR_RED
+            bool "Red"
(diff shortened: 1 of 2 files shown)


thekurtovic:
WIP
I've had problems in the past when needing to enable debug level logging, since this normally depends on IDF logging.
There tends to be so much debug printing that it causes my board to go into a boot loop. I've been modifying `NimBLELog.h` locally to address this. 
Also since I use ANSI color codes, I find it's difficult to distinguish my application logging vs NimBLE logging. I thought allowing for the colors to be customized would be helpful for parsing log output.

h2zero:
At first glance I think it's an interesting idea. I was thinking about overhauling the logging as well, I've removed quite a bit of it but better filtering is needed and the custom color idea is great. I'll look at this more in the next days but I have quite a bit on my agenda and want to publish a version 2 release before Christmas, so this might have to come after that.

h2zero:
@thekurtovic Unfortunately this breaks the log filtering capability.
For your use have you tried the filtering options? i.e. :
```
  esp_log_level_set("*", ESP_LOG_ERROR);
  esp_log_level_set("NimBLEDevice", ESP_LOG_DEBUG);
  esp_log_level_set("NimBLEAdvertisedDevice", ESP_LOG_DEBUG);
  esp_log_level_set("NimBLEScan", ESP_LOG_DEBUG);
```

thekurtovic:
This was meant to be an option which doesn't rely on esp logging, the loss of filtering would be a trade-off. Also wasn't aware you could even do that.
This seems pretty niche so I'll just close and make a patch file for my project.

h2zero:
Unfortunately to change the colors we need to bypass the log filtering and that wouldn't be great. Here is some more info though, https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/log.html.

thekurtovic:
The color part can still work with filtering like this.
https://github.com/h2zero/esp-nimble-cpp/commit/a34ed8fd00f1a40bd2271cb68244a9f8150e8965

h2zero:
@thekurtovic That looks good, have you tested it?

thekurtovic:
Yes, tested on esp32-s3. 
I've set log level to debug then tested with `esp_log_level_set("*", ESP_LOG_WARN);`. 
Then tested setting nimble log levels like this.
```
esp_log_level_set("*", ESP_LOG_WARN);
esp_log_level_set("NimBLEScan", ESP_LOG_DEBUG);
esp_log_level_set("NimBLEClient", ESP_LOG_DEBUG);
esp_log_level_set("NimBLEDevice", ESP_LOG_DEBUG);
esp_log_level_set("NimBLEAdvertisedDevice", ESP_LOG_DEBUG);
```

h2zero:
Perfect, please open a new PR or reopen this one and I will look to test/merge it after this release is done. Thanks for all the contributions!

PR #267, opened 2024-12-14
NimBLELog allow custom log level colors.

thekurtovic:
https://github.com/h2zero/esp-nimble-cpp/pull/256


h2zero:
Can the macros be renamed to use `NIMBLE_LOG_CPP` prefix so we can avoid potential issues with the actual nimble stack config macros?

h2zero:
Thanks!

PR #272, opened 2024-12-20
feat(char): Update characteristic to check properties/encryption

finger563:
* Update to check indication/notification properties in `NimBLECharacteristic::setValue`
* Update to check encrytpion of connection in `NimBLECharacteristic::setValue`

As best I can tell this is not needed, but I figured I'd open the PR since I wrote the code as a copy-over from the older version of this code. Feel free to reject this PR if its superfluous.


h2zero:
Thanks, I don't believe this to be needed though as you suspected. The client will not be able to subscribe without encrypting first so there shouldn't be a need. And in the case of custom notification I'm sure the app that calls this will know that's what it wants to do.

PR #274, opened 2024-12-21
Refactor NimBLEAdvertisementData/NimBLEExtAdvertising

thekurtovic:
* NimBLEUtils added method to get UUID type from size.
* NimBLEUtils added method to get service data UUID type from size.
* NimBLEUUID added method to get location within payload.
* NimBLEUUID added method as wrapper for NimBLEUtils::getUUIDType().

### DIFF
--- a/src/NimBLEUUID.cpp
+++ b/src/NimBLEUUID.cpp
@@ -173,6 +173,32 @@ const ble_uuid_t* NimBLEUUID::getBase() const {
 } // getBase
 
 /**
+ * @brief Get the UUID type based on size in bits.
+ * @return BLE_HS_ADV_TYPE_COMP_UUIDS16/32/128 or BLE_HS_ADV_TYPE_INCOMP_UUIDS16/32/128
+ *         for valid input size, else 0.
+ */
+uint8_t NimBLEUUID::getType() const {
+    return NimBLEUtils::getUUIDType(bitSize(), true);
+} // getType
+
+/**
+ * @brief Get the location of the UUID within a payload.
+ * @param [in] dataLoc The data location.
+ * @param [in] payload The payload to search through.
+ * @return >= 0 if found, else -1.
+ */
+int NimBLEUUID::getLoc(int dataLoc, const std::vector<uint8_t> payload) const {
+    uint8_t bytes = bitSize() >> 3;
+    const size_t size = payload.size();
+    for (size_t i = dataLoc + 2; i < size; i += bytes) {
+        if (memcmp(&payload[i], getValue(), bytes) == 0) {
+            return i;
+        }
+    }
+    return -1;
+} // getLoc
+
+/**
  * @brief Compare a UUID against this UUID.
  *
  * @param [in] uuid The UUID to compare against.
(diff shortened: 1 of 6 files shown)


h2zero:
Thanks! Good idea, I think changing the name to `getUUIDType` might be more appropriate?

thekurtovic:
Would it make more sense if I put the method inside `NimBLEUUID` itself?

src/NimBLEUtils.cpp
```diff
@@ -572,4 +572,42 @@ NimBLEAddress NimBLEUtils::generateAddr(bool nrpa) {
...
+            return BLE_HS_ADV_TYPE_SVC_DATA_UUID16;
+        case 4:
+            return BLE_HS_ADV_TYPE_SVC_DATA_UUID32;
+        case 16:
+            return BLE_HS_ADV_TYPE_SVC_DATA_UUID128;
+        default:
+            return 0;
+    }
+} // getServiceType
+
```
h2zero, line 612: Since the return value of these only relates to and is only used for advertising I feel they would be better placed in the advertising classes.

h2zero: I think we should also rename `getServiceType` to `getServiceDataType`

src/NimBLEUUID.cpp
```diff
@@ -172,6 +172,32 @@ const ble_uuid_t* NimBLEUUID::getBase() const {
...
+    uint8_t bytes = bitSize() >> 3;
+    const size_t size = payload.size();
+    for (size_t i = dataLoc + 2; i < size; i += bytes) {
+        if (memcmp(&payload[i], getValue(), bytes) == 0) {
+            return i;
+        }
+    }
+    return -1;
+} // getLoc
+
```
h2zero, line 200: I'm not sure these belong in the uuid class since the type returned is an advertisement uuid type and not the actual uuid type and `getLoc` probably isnt useful outside of advertising data. Not sure what the use case is as a public API.

h2zero (commented):
I like the consolidation idea, I think another approach may be better though.

I'm thinking these utility functions belong more in the advertising data class and that extended advertising should use or inherit from the class to keep the code all in one place to reduce the duplication.

thekurtovic:
I had considered the same, and thought that would be more appropriate but didn't want to go down that route in case you had any objections to a big change.
I can close this PR if you'd prefer to refactor with inheritance.

src/NimBLEUUID.cpp
thekurtovic: Admittedly this one didn't really make sense to me, but was purely in the pursuit of avoiding duplicate code. I put it in `NimBLEUUID` since the snippet was accessing accessing a UUID object originally.
Has no business being public really.

src/NimBLEUtils.cpp
thekurtovic: Good points

h2zero:
Since this isn't really essential at the moment I think it best to refactor with the larger scale approach so the changes make more sense to the overall classes. Not sure yet if inheritance is the best approach or if the extended advertising class should just have a member variable of  `NimBLEAdvertisingData`.

Issue #277, opened 2024-12-25
Heap corruption


thekurtovic:
Tested with esp32-s3 on 5.3.2.241210.
If I run an example project with `CONFIG_HEAP_POISONING_COMPREHENSIVE` enabled, I consistently get crashes due to memory.
`CORRUPT HEAP: Invalid data at 0x3fcbfaec. Expected 0xfefefefe got 0xfefefffe` 
<hr>
According to the docs

> If a call to [heap_caps_malloc()](https://docs.espressif.com/projects/esp-idf/en/v5.3.2/esp32s3/api-reference/system/mem_alloc.html#_CPPv416heap_caps_malloc6size_t8uint32_t) or [heap_caps_realloc()](https://docs.espressif.com/projects/esp-idf/en/v5.3.2/esp32s3/api-reference/system/mem_alloc.html#_CPPv417heap_caps_reallocPv6size_t8uint32_t) causes a crash because it was expected to find the pattern 0xFEFEFEFE in free memory and a different pattern was found, it indicates that the app has a use-after-free bug where it is writing to memory that has already been freed.
<hr>

The code is basically the [NimBLE_Client](https://github.com/h2zero/esp-nimble-cpp/blob/master/examples/NimBLE_Client/main/main.cpp) example with an `onResult` handler like so.
```
void onResult(const NimBLEAdvertisedDevice* adv) override {
        LOG_D("%s", advDevice.getAddress().toString().c_str());
        advDevice = *adv;
        pScan->stop();
        doConnect = true;
    }
```

Appears to have been introduced by 21513860577a6ebcac319d6ec0e190cf91662f65.
I can upload an example project if it ends up not being reproducible for you.

h2zero:
Easy fix, do t call scan stop

Issue #242, opened 2024-11-26
NimBLEClient::connect not using expected peer address


thekurtovic:
Occasionally I'm getting connection failures seemingly due to the target peer address being incorrectly set.
I've only noticed it when reusing client objects, but I always manually set the peer address. Added some extra log messages in NimBLEClient to track the address along the way.
Testing on master with esp32-s3.

```
I [142994][main.cpp:257] BLE adv (-57 db, phy 1 0) Name: , Address: e3:e4:5b:75:ad:36
D [143016][main.cpp:280] Reusing client, set addr->e3:e4:5b:75:ad:36
W (143026) NimBLEClient: setPeerAddress->35:3a:34:65:3a:33
W (143036) NimBLEClient: connect set addr->35:3a:34:65:3a:33
E (143037) NimBLEClient: Failed to connect to 35:3a:34:65:3a:33, rc=3;
E [143047][main:293] Failed to connect
```
```
void ScanCallbacks::onResult(const NimBLEAdvertisedDevice *adv)
{
    ....
    auto client = NimBLEDevice::getDisconnectedClient();
    if (!client) {
        client = NimBLEDevice::createClient(adv->getAddress());
        if (!client) {
            // TODO state error handling..
            LOG_E("Failed to create client");
            return;
        }
    } else {
        LOG_D("Reusing client, set addr->%s", adv->getAddress().toString().c_str());
        client->setPeerAddress(adv->getAddress());
    }
  
    client->setClientCallbacks(&clientCB, false);
    client->setConnectionParams(6, 6, 0, 400);
    if (!client->connect(true, true, true))
        LOG_E("Failed to connect");
    ....
}
```

Adding `const NimBLEAddress address = adv->getAddress();` at the start of `onResult` seems to avoid the issue, but I'll let it run for the day to confirm.

h2zero:
@thekurtovic I'm not able to reproduce this, I did find another issue though. I copied the code you provided into my onResult callback and restarted the server device multiple times and could reconnect each time without issue.

thekurtovic:
Thanks for checking. I believe this only started the other day, I must have done something stupid in my code.
I'll roll things back on my end and double check with a simplified `onResult` handler.
Closing for now assuming user error.

thekurtovic:
[...]

PR #278, opened 2024-12-25
NimBLEScan increment m_callbackSent before callback.

thekurtovic:
Fixes heap corruption observed on CONFIG_HEAP_POISONING_COMPREHENSIVE.
Resolves https://github.com/h2zero/esp-nimble-cpp/issues/277

### DIFF
--- a/src/NimBLEScan.cpp
+++ b/src/NimBLEScan.cpp
@@ -119,19 +119,19 @@ int NimBLEScan::handleGapEvent(ble_gap_event* event, void* arg) {
             }
 
             if (!advertisedDevice->m_callbackSent) {
-                pScan->m_pScanCallbacks->onDiscovered(advertisedDevice);
                 advertisedDevice->m_callbackSent++;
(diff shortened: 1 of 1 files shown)


thekurtovic:
I have no idea what's going on here. Checked for other references on `m_callbackSent` but can't see anything which would require the increment to happen before the callback.
I originally thought calling `pScan->stop();` explicitly was causing issues (like https://github.com/h2zero/esp-nimble-cpp/issues/242), as it appeared as though letting `NimBLEClient::connect()` implicitly stop the scan would avoid the issue, but it seems to not be true anymore.

h2zero (commented):
LGTM, thanks!

thekurtovic:
It makes sense now.
Incrementing after meant that there was a risk of a dangling pointer since the scan could be stopped within the `onResult` callback. Best to always have the callback "last" when possible.

h2zero:
Yes, I'm not sure why it was changed but that would be the reason for the crashing.

PR #280, opened 2025-01-03
Add checks in case NIMBLE_CPP_DEBUG_ASSERT is not defined.

thekurtovic:


thekurtovic:
I haven't personally experienced any issues in these spots, but I noticed they were only checked by the assert macro which is disabled by default.
Could just modify the macro itself instead, might be cleaner if including `NimBLELog` into `nimconfig.h` is acceptable.

src/NimBLEAttValue.cpp
```diff
@@ -57,6 +62,10 @@ NimBLEAttValue::~NimBLEAttValue() {
 // Move assignment operator implementation.
 NimBLEAttValue& NimBLEAttValue::operator=(NimBLEAttValue&& source) {
     if (this != &source) {
+        if (m_attr_value == nullptr) {
+            NIMBLE_LOGE(LOG_TAG, "Move assign nullptr");
+            return *this;
+        }
```
h2zero, line 68: I don't think this is necessary as `m_attr_value` should not be nullptr unless the constructor failed to allocate.

src/NimBLEDevice.cpp
```diff
@@ -1231,7 +1231,7 @@ std::string NimBLEDevice::toString() {
  * @param [in] line The line number where the assert occurred.
  */
 void nimble_cpp_assert(const char* file, unsigned line) {
-    console_printf("Assertion failed at %s:%u\n", file, line);
+    NIMBLE_LOGE("", "Assertion failed at %s:%u\n", file, line);
```
h2zero, line 1234: Not sure about this one, if the log filtering is on then abort will be called without any information.

h2zero (commented):
Looks good, just a couple of comments.

src/NimBLEAttValue.cpp
thekurtovic: I forget that `free(nullptr)` is okay every now and then.

src/NimBLEDevice.cpp
thekurtovic: Oh right, I keep forgetting about that feature. I just wanted to have it be red for users with log colors enabled.
I changed it like so, is it acceptable?
`console_printf(NIMBLE_LOG_FORMAT(E, "Assertion failed at %s:%u"), 0, "", file, line);`

src/NimBLEDevice.cpp
```diff
@@ -1231,7 +1231,7 @@ std::string NimBLEDevice::toString() {
  * @param [in] line The line number where the assert occurred.
  */
 void nimble_cpp_assert(const char* file, unsigned line) {
-    console_printf("Assertion failed at %s:%u\n", file, line);
+    console_printf(NIMBLE_LOG_FORMAT(E, "Assertion failed at %s:%u"), 0, "", file, line);
```
h2zero, line 1234: I think we should not change this as it would fail in the Arduino environment.

src/NimBLEAttValue.h
```diff
@@ -325,6 +325,10 @@ class NimBLEAttValue {
     /** @brief Subscript operator */
     uint8_t operator[](int pos) const {
         NIMBLE_CPP_DEBUG_ASSERT(pos < m_attr_len);
+        if (pos >= m_attr_len) {
+            NIMBLE_LOGE("NimBLEAttValue", "pos >= len, pos=%u, len=%u", pos, m_attr_len);
+            return 0;
+        }
```
h2zero, line 331: Maybe we should move this function definition to the cpp file?

h2zero (changes requested):
Looks good, just a couple of things to tweak.

h2zero (approved):
LGTM Thanks!

PR #281, opened 2025-01-04
Add explicit keyword to constructors

thekurtovic:
Resolves cppcheck warnings noExplicitConstructor, and useInitializationList.

### DIFF
--- a/src/NimBLE2904.h
+++ b/src/NimBLE2904.h
@@ -38,7 +38,7 @@ struct NimBLE2904Data {
  */
 class NimBLE2904 : public NimBLEDescriptor {
   public:
-    NimBLE2904(NimBLECharacteristic* pChr = nullptr);
+    explicit NimBLE2904(NimBLECharacteristic* pChr = nullptr);
     static const uint8_t FORMAT_BOOLEAN   = 1;
     static const uint8_t FORMAT_UINT2     = 2;
     static const uint8_t FORMAT_UINT4     = 3;
--- a/src/NimBLEAddress.h
+++ b/src/NimBLEAddress.h
@@ -44,10 +44,10 @@ class NimBLEAddress : private ble_addr_t {
      * @brief Create a blank address, i.e. 00:00:00:00:00:00, type 0.
      */
     NimBLEAddress() = default;
-    NimBLEAddress(const ble_addr_t address);
     NimBLEAddress(const uint8_t address[BLE_DEV_ADDR_LEN], uint8_t type);
     NimBLEAddress(const std::string& stringAddress, uint8_t type);
     NimBLEAddress(const uint64_t& address, uint8_t type);
+    explicit NimBLEAddress(const ble_addr_t address);
 
     bool                 isRpa() const;
     bool                 isNrpa() const;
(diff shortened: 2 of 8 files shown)


thekurtovic:
There's one warning I wasn't able to resolve.
`NimBLEUUID(const ble_uuid_any_t& uuid);`

If I make that constructor explicit then I get this compile error.
<details>
<summary>Log</summary>

```
managed_components/esp-nimble-cpp/src/NimBLERemoteValueAttribute.h: In constructor 'NimBLERemoteValueAttribute::NimBLERemoteValueAttribute(const ble_uuid_any_t&, uint16_t)':
managed_components/esp-nimble-cpp/src/NimBLERemoteValueAttribute.h:189:107: error: no matching function for call to 'NimBLEAttribute::NimBLEAttribute(const ble_uuid_any_t&, uint16_t&)'
  189 |     NimBLERemoteValueAttribute(const ble_uuid_any_t& uuid, uint16_t handle) : NimBLEAttribute(uuid, handle) {}
      |                                                                                                           ^
In file included from managed_components/esp-nimble-cpp/src/NimBLERemoteService.h:24,
                 from managed_components/esp-nimble-cpp/src/NimBLEDevice.h:261:
managed_components/esp-nimble-cpp/src/NimBLEAttribute.h:48:5: note: candidate: 'NimBLEAttribute::NimBLEAttribute(const NimBLEUUID&, uint16_t)'
   48 |     NimBLEAttribute(const NimBLEUUID& uuid, uint16_t handle) : m_uuid{uuid}, m_handle{handle} {}
      |     ^~~~~~~~~~~~~~~
managed_components/esp-nimble-cpp/src/NimBLEAttribute.h:48:39: note:   no known conversion for argument 1 from 'const ble_uuid_any_t' to 'const NimBLEUUID&'
   48 |     NimBLEAttribute(const NimBLEUUID& uuid, uint16_t handle) : m_uuid{uuid}, m_handle{handle} {}
      |                     ~~~~~~~~~~~~~~~~~~^~~~
managed_components/esp-nimble-cpp/src/NimBLEAttribute.h:29:7: note: candidate: 'constexpr NimBLEAttribute::NimBLEAttribute(const NimBLEAttribute&)'
   29 | class NimBLEAttribute {
```

</details>

<hr>

Also `NimBLEHIDDevice` needs to be updated to be compliant with this change.
For example: `m_deviceInfoSvc = server->createService(deviceInfoSvcUuid);`

Do you think it's better to just create a new constructor which can take a uint16_t, or change such instances like this?
`m_deviceInfoSvc = server->createService(NimBLEUUID(deviceInfoSvcUuid));`

h2zero:
I think `m_deviceInfoSvc = server->createService(NimBLEUUID(deviceInfoSvcUuid));` is probably fine here as it's an internally managed class so application code isn't affected, still quite readable as well.

thekurtovic:
Would this be considered a breaking change? Any applications making use of these implicit conversions will fail to compile, though the changes should be easy to make.
I'll make a separate PR for `NimBLEHIDDevice`.

h2zero:
It might be, depending on the application.

I would suggest adding the HID changes to this PR so we can test the builds at least. Maybe add some more build tests to find where it may fail so it can be documented?

PR #282, opened 2025-01-06
NimBLEAddress::toString allow formatting to be configured

thekurtovic:
* Added Kconfig bool option `NIMBLE_CPP_ADDR_FMT_UPPERCASE`
  If enabled, allows address to be printed in uppercase letters. (Default: No)
* Added Kconfig bool option `NIMBLE_CPP_ADDR_FMT_INCLUDE_DELIMITER`
  If enabled, allows address to be printed with colon characters included. (Default: Yes)

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -157,6 +157,20 @@ config NIMBLE_CPP_ENABLE_ADVERTISEMENT_TYPE_TEXT
         while scanning as text messages in the debug log.
         This will use approximately 250 bytes of flash memory.
 
+config NIMBLE_CPP_ADDR_FMT_INCLUDE_DELIMITER
+    bool "Show colon characters when printing address."
+    default "y"
+    help
+        Enabling this option will format MAC addresses with
+        colon characters included when printing.
+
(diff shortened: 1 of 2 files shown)


thekurtovic:
Use case: I usually have addresses printed without colons for the majority of my application, and being able to have some control over address printing format would be useful when ctrl+F searching through a log file. Also avoids having to add code in the application to manage it.

h2zero:
Nice! I could see this being helpful for a few folks. I wonder though if making this runtime selectable rather and a config option would be a better approach by adding parameters to specify the output to the `toString` method?

h2zero:
Right... I forgot about that operator. I that case lets go with the config options.

thekurtovic:
Did you mean Arduino? Platformio does have access to kconfig. Maybe I'm misunderstanding the problem

h2zero:
when using platformio and this repo in lib_deps with espidf framework it doesn't read the component config from the library so the values need to be specified in build_flags unfortunately.

h2zero:
Yep, that will work, thanks!

PR #287, opened 2025-01-11
refactor(RemoteChar): Reduce nesting

thekurtovic:
* Renamed `desc_filter_t` to `NimBLEDescriptorFilter`
* Added `NimBLERemoteDescriptor` pointer to `NimBLEDescriptorFilter`
* `retrieveDescriptors` changed to take `NimBLEDescriptorFilter` pointer
* General cleanup

### DIFF
--- a/src/NimBLERemoteCharacteristic.cpp
+++ b/src/NimBLERemoteCharacteristic.cpp
@@ -27,17 +27,18 @@
 
 # include <climits>
 
-typedef struct {
-    const NimBLEUUID* uuid;
-    void*             task_data;
-} desc_filter_t;
+struct NimBLEDescriptorFilter {
+    NimBLERemoteDescriptor* dsc;
+    const NimBLEUUID*       uuid;
+    void*                   taskData;
+};
 
 static const char* LOG_TAG = "NimBLERemoteCharacteristic";
 
 /**
  * @brief Constructor.
  * @param [in] svc A pointer to the service this characteristic belongs to.
- * @param [in] ble_gatt_chr struct defined as:
+ * @param [in] chr struct defined as:
  *  struct ble_gatt_chr {
  *      uint16_t def_handle;
  *      uint16_t val_handle;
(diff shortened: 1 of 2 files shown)


thekurtovic:
@h2zero Not sure if I'm missing something obvious, but where does the task from `retrieveDescriptors` get released if `rc == 0` in `descriptorDiscCB`? 
It appears appears to only get released if `rc != 0` after https://github.com/h2zero/esp-nimble-cpp/commit/0b543d9f627b66bb4ae336b9fec3531f47c4d1f5. Is it being done by `NimBLERemoteService::characteristicDiscCB` somehow?

src/NimBLERemoteCharacteristic.cpp
```diff
@@ -72,35 +72,31 @@ int NimBLERemoteCharacteristic::descriptorDiscCB(
...
+    if (!rc && (!uuidFilter || !ble_uuid_cmp(uuidFilter->getBase(), &dsc->uuid.u))) {
+        rc = BLE_HS_EDONE; // Indicate that descriptor was found.
         pChr->m_vDescriptors.push_back(new NimBLERemoteDescriptor(pChr, dsc));
-        return 0;
+    }
+    else if (rc != 0) {
+        NimBLEUtils::taskRelease(*pTaskData, rc);
     }
 
-    NimBLEUtils::taskRelease(*pTaskData, rc);
```
h2zero, line 92: I don't think this will work right, if rc == 0 it means a descriptor was found, then we need to check if it is the one we are looking for and add it to the vector if it was, then release the task and return BLE_HS_EDONE to stop the results from the stack. If no filter is provided we keep adding the results to the vector until the error->status is not 0 with BLE_HS_EDONE being no more results to come.

h2zero (commented):
Nice cleanup, thanks! Need to make some changes to retrieveDescriptors though.

src/NimBLERemoteCharacteristic.cpp
thekurtovic: I think it now matches your description (hopefully). Still unsure of the task releasing here. It now will release on `rc == 0` but not for the check above `if (pChr->getHandle() != chrHandle)`.

thekurtovic:
Only part I'm unsure about right now is whether
```
if (rc == BLE_HS_EDONE) {
    NimBLEUtils::taskRelease(*pTaskData, rc);
}
```
should be
```
if (rc != 0) {
    NimBLEUtils::taskRelease(*pTaskData, rc);
}
```

src/NimBLERemoteCharacteristic.cpp
```diff
@@ -63,44 +63,41 @@ NimBLERemoteCharacteristic::~NimBLERemoteCharacteristic() {
...
+    auto       pTaskData = (NimBLETaskData*)filter->task_data;
+    const auto pChr      = (NimBLERemoteCharacteristic*)pTaskData->m_pInstance;
+    const auto uuid      = filter->uuid; // UUID to filter for
 
-    if (pChr->getHandle() != chr_val_handle) {
-        return 0; // Descriptor not for this characteristic
+    // Error/done, wrong characteristic, or no-match descriptor
+    if (rc != 0 || pChr->getHandle() != chrHandle
+      || (uuid && ble_uuid_cmp(uuid->getBase(), &dsc->uuid.u))) {
+        goto Done;
```
h2zero, line 78: If we have specified a UUID filter and rc == 0 we should return 0 so the next descriptor is fetched, until error or BLE_HS_EDONE.

src/NimBLERemoteCharacteristic.cpp
```diff
@@ -63,44 +63,41 @@ NimBLERemoteCharacteristic::~NimBLERemoteCharacteristic() {
...
-                rc = BLE_HS_EDONE; // Found the descriptor, stop the search
-            } else {
-                return 0; // Not the descriptor we are looking for
-            }
-        }
-
-        pChr->m_vDescriptors.push_back(new NimBLERemoteDescriptor(pChr, dsc));
-        return 0;
+Done:
+    if (rc == BLE_HS_EDONE) {
```
h2zero, line 86: we also need to release the task on error, BLE_HS_EDONE is only sent when all the descriptors are discovered.

h2zero (commented):
Almost there 👍

src/NimBLERemoteCharacteristic.cpp
thekurtovic: It should cover that case.
Evaluates to `if (false || n/a || (true && x))` where x is the `ble_uuid_cmp` call. If the compare is non-zero then `rc == 0` is returned. If the compare is zero then `rc == BLE_HS_EDONE` is returned.

h2zero: You are correct, my bad, I read this wrong.

src/NimBLERemoteCharacteristic.cpp
```diff
@@ -63,44 +63,41 @@ NimBLERemoteCharacteristic::~NimBLERemoteCharacteristic() {
...
-            } else {
-                return 0; // Not the descriptor we are looking for
-            }
-        }
-
-        pChr->m_vDescriptors.push_back(new NimBLERemoteDescriptor(pChr, dsc));
-        return 0;
+Done:
+    if (rc != 0) {
+        NimBLEUtils::taskRelease(*pTaskData, rc);
```
h2zero, line 87: Can we move the debug message below inside this if block so it only prints when the discovery is complete?

src/NimBLERemoteCharacteristic.cpp
```diff
@@ -63,44 +63,41 @@ NimBLERemoteCharacteristic::~NimBLERemoteCharacteristic() {
...
 
 /**
  * @brief Populate the descriptors (if any) for this characteristic.
- * @param [in] the end handle of the characteristic, or the service, whichever comes first.
+ * @param [in] uuidFilter The end handle of the characteristic, or the service, whichever comes first.
+ * @param [out] out Pointer to get the descriptor if found.
+ * @return True if successfully retrieved, success = 0 or BLE_HS_EDONE.
  */
-bool NimBLERemoteCharacteristic::retrieveDescriptors(const NimBLEUUID* uuidFilter) const {
+bool NimBLERemoteCharacteristic::retrieveDescriptors(const NimBLEUUID* uuidFilter, NimBLERemoteDescriptor* out) const {
```
h2zero, line 99: Maybe instead of adding a parameter we should just add a `NimBLERemoteDescriptor*` to `desc_filter_t` and pass a pointer to the filter as the parameter to this. Would require creating the instance of the filter in `getDescriptor` instead of course?

src/NimBLERemoteCharacteristic.cpp
```diff
@@ -118,6 +115,9 @@ bool NimBLERemoteCharacteristic::retrieveDescriptors(const NimBLEUUID* uuidFilte
     rc = taskData.m_flags;
     if (rc == 0 || rc == BLE_HS_EDONE) {
         NIMBLE_LOGD(LOG_TAG, "<< retrieveDescriptors(): found %d descriptors.", m_vDescriptors.size());
+        if (out && rc) {
+            out = m_vDescriptors.back();
+        }
```
h2zero, line 120: I think we can remove the test for `rc == 0` from this outer block here and just check for `BLE_HS_EDONE`, 0 should not be returned in `rc` for this function. This block could also be removed if we make the change to `desc_filter_t` above and set the pointer to the descriptor in the callback.

h2zero (commented):
Couple of thoughts to extend the changes, let me know what you think.

src/NimBLERemoteCharacteristic.cpp
thekurtovic: I've left setting the pointer within `retrieveDescriptors` since it already has an if-check for `BLE_HS_EDONE`.

thekurtovic:
Moved things around in `descriptorDiscCB` to avoid `goto`.
Had to change the definition of `desc_filter_t` to make the forward declaration work.

src/NimBLERemoteCharacteristic.cpp
```diff
@@ -27,17 +27,18 @@
...
 
-typedef struct {
-    const NimBLEUUID* uuid;
-    void*             task_data;
-} desc_filter_t;
+struct desc_filter_t {
+    NimBLERemoteDescriptor* dsc;
+    const NimBLEUUID*       uuid;
+    void*                   taskData;
+};
```
h2zero, line 34: Since this is no longer a typedef can you rename it to `NimBLEDescriptorFilter` please?

src/NimBLERemoteCharacteristic.cpp
```diff
@@ -132,51 +126,36 @@ bool NimBLERemoteCharacteristic::retrieveDescriptors(const NimBLEUUID* uuidFilte
...
-    NimBLERemoteDescriptor* pDsc      = nullptr;
-    size_t                  prev_size = m_vDescriptors.size();
+    NimBLETaskData taskData(const_cast<NimBLERemoteCharacteristic*>(this));
+    desc_filter_t  filter = {nullptr, &uuid, &taskData};
+    NimBLEUUID     uuidTmp;
 
     for (const auto& it : m_vDescriptors) {
         if (it->getUUID() == uuid) {
-            pDsc = it;
+            filter.dsc = it;
```
h2zero, line 135: Since we are in here, could you change the variable name from `it` to `dsc` since that is not an iterator.

h2zero (commented):
Sorry to be a pain, just a couple minor items.

h2zero (commented):
LGTM! Thanks!

h2zero:
@thekurtovic Could you please rebase this? Master added a check that should remain.

thekurtovic:
Still contains these requested changes
https://github.com/h2zero/esp-nimble-cpp/pull/287/#discussion_r1913579319

h2zero:
Might need to try again, I fixed the build issue but there is a merge conflict.

thekurtovic:
If I resolve the conflict then it adds an additional commit to this PR. Am I supposed to squash everything into my main commit?

h2zero:
I can squash the commits after when merging.

h2zero:
Thanks, I rebased it and pushed back to your branch.

thekurtovic:
Thanks, I was just about to try something myself. 

Just fore future reference, would this have worked?
```
git checkout refactor
git checkout -b temp-squash-branch
...squash commits back my original commit...
git push --force-with-lease origin refactor
```

h2zero:
All I did was:
`git rebase master`, 
fixed the conflicts, 
`git add .`
`git rebase --continue`
`git push -f origin refactor`

thekurtovic:
Thanks. I was trying to rebase my branch after fixing the conflict but the rebase kept showing me commits from master only for whatever reason. I thought master was being merged into the branch opposed to the other way around. I need to lookup a git tutorial

PR #289, opened 2025-01-12
Fix race condition in NimBLEScan::clearResults.

h2zero:
If clear results is called from more than one task a race condition exists that may delete the same advertisedDevice twice. This prevents this by swapping with an empty vector and testing for empty before freeing the resources.


thekurtovic:
Does this apply to other areas in the codebase which have a similar clear/delete method?
https://github.com/h2zero/esp-nimble-cpp/blob/master/src/NimBLEClient.cpp#L102-L109

h2zero:
Shouldn't be an issue here unless deleting the client instance from multiple tasks, which would be quite odd.

PR #291, opened 2025-01-13
refactor(RemoteServ): Reduce nesting

thekurtovic:
* Adds parameter out to retrieveCharacteristics, default valuel works as the original method.
    * out: if not nullptr, will allow caller to obtain the characteristic
* General cleanup

### DIFF
--- a/src/NimBLERemoteService.cpp
+++ b/src/NimBLERemoteService.cpp
@@ -76,46 +76,31 @@ NimBLERemoteCharacteristic* NimBLERemoteService::getCharacteristic(const char* u
  */
 NimBLERemoteCharacteristic* NimBLERemoteService::getCharacteristic(const NimBLEUUID& uuid) const {
     NIMBLE_LOGD(LOG_TAG, ">> getCharacteristic: uuid: %s", uuid.toString().c_str());
-    NimBLERemoteCharacteristic* pChar     = nullptr;
-    size_t                      prev_size = m_vChars.size();
+    NimBLERemoteCharacteristic* pChar = nullptr;
+    NimBLEUUID                  uuidTmp;
 
-    for (const auto& it : m_vChars) {
-        if (it->getUUID() == uuid) {
-            pChar = it;
+    for (const auto& chr : m_vChars) {
+        if (chr->getUUID() == uuid) {
+            pChar = chr;
             goto Done;
         }
     }
 
-    if (retrieveCharacteristics(&uuid)) {
-        if (m_vChars.size() > prev_size) {
-            pChar = m_vChars.back();
-            goto Done;
-        }
+    if (!retrieveCharacteristics(&uuid, pChar) || pChar) {
+        goto Done;
+    }
 
-        // If the request was successful but 16/32 bit uuid not found
-        // try again with the 128 bit uuid.
-        if (uuid.bitSize() == BLE_UUID_TYPE_16 || uuid.bitSize() == BLE_UUID_TYPE_32) {
-            NimBLEUUID uuid128(uuid);
-            uuid128.to128();
-            if (retrieveCharacteristics(&uuid128)) {
-                if (m_vChars.size() > prev_size) {
-                    pChar = m_vChars.back();
-                }
-            }
-        } else {
-            // If the request was successful but the 128 bit uuid not found
-            // try again with the 16 bit uuid.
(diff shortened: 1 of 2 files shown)


thekurtovic:
Same treatment as previous PR.
`NimBLEClient` also has a similar section of code. I was thinking of making a class which the other 3 classes will be derived from, then move the retrieve function into the new class and make it a template.

Might have to roll back to `bool retrieveDescriptors(const NimBLEUUID* uuidFilter = nullptr, NimBLERemoteDescriptor* out = nullptr) const;` in `NimBLERemoteCharacteristic` to make it work though.

h2zero:
Looks good and I agree with the consolidation thought. Not sure if creating/inheriting from a class for this purpose is the way to go because the client class would also need to inherit from it. Maybe a utility function instead?

thekurtovic:
Sure, that works too.

thekurtovic:
Does `retrieveCharacteristics` need a check similar to https://github.com/h2zero/esp-nimble-cpp/commit/8158a16fbf3d716fd2dd16f6f8990c12265aefe7?

h2zero:
Probably not, unless it's an empty service, which would be weird. Wouldn't be the worst idea to add it anyway though.

thekurtovic:
Alt Text

src/NimBLERemoteService.cpp
```diff
@@ -143,77 +127,68 @@ const std::vector<NimBLERemoteCharacteristic*>& NimBLERemoteService::getCharacte
...
-        NIMBLE_LOGD(LOG_TAG, "<< retrieveCharacteristics()");
-        return true;
+    if (rc != BLE_HS_EDONE) {
+        NIMBLE_LOGE(LOG_TAG, "<< retrieveCharacteristics(): failed: rc=%d %s", rc, NimBLEUtils::returnCodeToString(rc));
+        return false;
     }
 
-    NIMBLE_LOGE(LOG_TAG, "<< retrieveCharacteristics() rc=%d %s", rc, NimBLEUtils::returnCodeToString(rc));
-    return false;
+    if (out) {
```
h2zero, line 187: 💯

h2zero: You have to love those mysterious compiler warnings when it comes to pointers, been there done that lol!

thekurtovic:
I introduced some minor changes since the previous successful build. Let me know if I should revert.
Since `characteristicDiscCB` doesn't mention the function name in the log messages, I shortened the name so that the multi-line function calls could be avoided in `retrieveCharacteristics`.
Also tried thinking of a more compact way to write `connHandle`, `endHandle`, etc. Settled on `hdleConn`, `hdlEnd`, but considered using h as a prefix to signify handles.

Btw when a PR gets merged, does it use the message from the OP or the message from the commit?

src/NimBLERemoteService.cpp
thekurtovic: I could have swore the consolidation PR omits the nullptr check and was able to pass.
Wasted so much time on this stupid error lol, especially since it was building fine on my end.
I'm assuming the build check uses more pedantic flags or something, would be nice to have a similar setup locally.

h2zero:
I'm not concerned with the naming, as long as it's somewhat consistent throughout the library and obvious/readable for casual observers as to what they represent, in this case the original naming suits but I'm happy to entertain changes.

>Btw when a PR gets merged, does it use the message from the OP or the message from the commit?

The title/message on github for each commit in a PR is what will be seen in git log when pulled by anyone else.

thekurtovic:
Some of the BLE jargon just adds so much bloat when written in full like characteristic, descriptor, etc.; which then causes the need to break function calls up with one parameter per line.

I definitely think some standardization should be established which will make things more concise while still having a clear meaning. Though that's probably a 3.0 discussion.

h2zero:
>Some of the BLE jargon just adds so much bloat when written in full like characteristic, descriptor, etc.; which then causes the need to break function calls up with one parameter per line.

Yeah, drives me crazy sometimes.

>I definitely think some standardization should be established which will make things more concise while still having a clear meaning. Though that's probably a 3.0 discussion.

Absolutely

>Good to know. Sometimes I mess up the message in the terminal and wasn't sure if I need to bother repushing

No promises lol, it comes down to the person merging, it can all be edited.

PR #292, opened 2025-01-14
refactor(Client): Reduce nesting

thekurtovic:
* Renames serviceDiscoveredCB to serviceDiscCB.
* Renames `exchangeMTUCb` to `exchangeMTUCB` to be consistent with codebase.
* Adds parameter out to retrieveServices, default value works as the original method.
    * out: if not nullptr, will allow caller to obtain the service
* General cleanup


h2zero:
Nice, all looks good to consolidate. I think we should communicate offsite,  if you're interested please contact me through the email in my profile.

PR #293, opened 2025-01-16
refactor: Consolidate attribute get code into NimBLEUtils

thekurtovic:


thekurtovic:
Basically includes the 3 refactors with slight changes to make the generic call work.
I wanted to move the code out of the header and use explicit template instantiation but haven't had any luck so far.

h2zero:
I need some time to test/review this further but it looks great so far.

h2zero:
I think we should move this outside of `NimBLEUtils` because this code should either belong to the client class or it's own remote attribute class so that it can be excluded when `CONFIG_BT_NIMBLE_ROLE_CENTRAL` is not defined.

thekurtovic:
I originally had the code like this, and had the template function in `NimBLESomething`.
`class NimBLEAttribute : public NimBLESomething`
`class NimBLEClient : public NimBLESomething`

Would that work? Can you can suggest a suitable name?

h2zero:
Instead of inheriting I think we should just add it to a header file `NimBLERemoteGattUtils.h`

h2zero:
The P4 build errors are not caused by this PR, upstream issue. Please add `fail-fast: false` here:      https://github.com/h2zero/esp-nimble-cpp/blob/8158a16fbf3d716fd2dd16f6f8990c12265aefe7/.github/workflows/build.yml#L15

PR #295, opened 2025-01-24
feat(Device): Implement store status handler

thekurtovic:


h2zero:
Nice, I've been meaning to do this for a long time, just never became a priority :) Thanks!

thekurtovic:
Updated documentation to clarify return codes.

h2zero:
The default callback should still call `ble_store_util_status_rr` to maintain the currently expected functionality.

PR #301, opened 2025-01-28
feat(Log): Add macros for conditional log print and rc handling

thekurtovic:
* Defined variants of the logging macros which print only if a condition is met.
    * These macros should be preferred over placing log statements inside a simple if-block (i.e. `rc != 0`).
* Defined a variant of the error log macro which takes in `rc` and implicitly appends `"; rc=%d %s"` to the format string in order to display return code and string representation via `NimBLEUtils::returnCodeToString`.
    * Prints only if `rc != 0`.

### DIFF
--- a/src/NimBLELog.h
+++ b/src/NimBLELog.h
@@ -110,16 +110,16 @@
 
 #   define NIMBLE_CPP_LOG_FORMAT(letter, format) NIMBLE_CPP_LOG_COLOR_##letter #letter " (%lu) %s: " format LOG_RESET_COLOR "\n"
 
-#   define NIMBLE_CPP_LOG_LEVEL_LOCAL(level, tag, format, ...)                                                                                                  		\
-      do {                                                                                                                                                          	\
-        if (level==ESP_LOG_ERROR)       { esp_log_write(ESP_LOG_ERROR, tag, NIMBLE_CPP_LOG_FORMAT(E, format), esp_log_timestamp(), tag __VA_OPT__(,) __VA_ARGS__); }	\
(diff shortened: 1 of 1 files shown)


thekurtovic:
There's a couple of spots where a log print will be surrounded by a simple if-check on a `rc` value.
```
if (rc != 0) {
    NIMBLE_LOGE(LOG_TAG, ...);
}
```

I thought it would be nice to have a macro for such cases to keep things a bit more clean going foward.
`NIMBLE_LOGE_IF(rc != 0, LOG_TAG, ...);`

What do you think?

h2zero:
Looks nice! We could probably add this part into the macro as well `status=%d`

thekurtovic:
Did a quick check and `rc=%d` seemed to be more common than `status=%d`.

h2zero (commented):
Looks good to me!

thekurtovic:
Hm seems like I did something wrong.
Commit history shows the one from this PR plus another one for "Merge pull request #301 from thekurtovic/feat-log-condition".

h2zero:
Ahh, no worries lol. I always rebase locally then push -f to the branch before merge but for others PR's I use the rebase and merge option and or squash

thekurtovic:
There's no way to resolve it through the github web ui is there? I see a "revert" button but it seems like that will just create more commits...
If you don't mind rebasing it at some point, I'd feel less dumb 🤐I regret not practicing on a test repo lol

h2zero:
A merge commit isn't going to hurt anything, it's just not my preference. Next time just do a squash to a single commit or however many you want and rebase to master from there. Then you can force push to your branch (carefully lol). After that there will be a straight line for the history. Oh and use the rebase merge option on GitHub, if the button goes grey then you know you need to do the above process.

PR #312, opened 2025-03-02
Fix ESP32-P4 compile when using Arduino as an ESP-IDF component

iranl:
btStarted() is not defined when SOC_BT_SUPPORTED is false


iranl:
Related fix for the Arduino framework has already been merged in https://github.com/espressif/arduino-esp32/pull/11035

h2zero:
Thanks!

PR #116, opened 2023-04-28
Add L2CAP service infrastructure

mickeyl:
Refers to #83.

This commit adds the ability to service BLE L2CAP connection-oriented-connections (COC).

It follows the GATT service and delegate patterns.

It has been tested successfully on ESP32C3 and ESP32C6 as peripherals and various iOS and MacOS devices as centrals.

NOTE: Please consider this a request-for-comments. It's probably lacking documentation and an example, but as this has been working very well for me, I thought I let it be reviewed by others.

PR #324, opened 2025-04-23
Introduce L2CAP infrastructure.

mickeyl:
L2CAP is the underlying technology powering GATT.
BLE 5 exposes L2CAP COC (Connection Oriented Channels) allowing a streaming API that leads to much higher throughputs than you can achieve with updating GATT characteristics.

The patch follows the established infrastructure very closely. The main components are:

- `NimBLEL2CAPChannel`, encapsulating an L2CAP COC.
- `NimBLEL2CAPServer`, encapsulating the L2CAP service.
- `Examples/L2CAP`, containing a client and a server application.

Apart from these, only minor adjustments to the existing code was necessary.

NOTE: The example programs point out a potential problem in the l2cap stack, probably deep inside Nimble, but we have to investigate that. If you are sending too many data, the stack silently drops it.

NOTE2: I forked the repository over two years ago and unfortunately didn't sync afterwards, so I had to do more merge work than expected to catch up with your changes. I hope everything is correct, but in order to reduce coming diffs, it's probably a good idea to pull this now.

This obsoletes https://github.com/h2zero/esp-nimble-cpp/pull/116


mickeyl:
Hmm… it fails to compile with ESP-IDF 4.4, probably due to the older C++ standard:

```
NimBLEL2CAPChannel.h:77:33: error: use of deleted function 'std::atomic<bool>::atomic(const std::atomic<bool>&)'
     std::atomic<bool> stalled = false;
                                 ^~~~~
In file included from /__w/esp-nimble-cpp/esp-nimble-cpp/components/esp-nimble-cpp/src/NimBLEL2CAPChannel.h:14,
                 from /__w/esp-nimble-cpp/esp-nimble-cpp/components/esp-nimble-cpp/src/NimBLEL2CAPChannel.cpp:4:
/opt/esp/tools/riscv32-esp-elf/esp-2021r2-patch5-8.4.0/riscv32-esp-elf/riscv32-esp-elf/include/c++/8.4.0/atomic:71:5: note: declared here
     atomic(const atomic&) = delete;
     ^~~~~~
```

How are we going to fix this correctly?

src/NimBLEL2CAPChannel.h
```diff
@@ -0,0 +1,115 @@
...
+    NimBLEL2CAPChannelCallbacks* callbacks;
+    uint8_t* receiveBuffer = nullptr; // buffers a full (local) MTU
+
+    // NimBLE memory pool
+    void* _coc_memory = nullptr;
+    struct os_mempool _coc_mempool;
+    struct os_mbuf_pool _coc_mbuf_pool;
+
+    // Runtime handling
+    std::atomic<bool> stalled = false;
```
h2zero, line 77: Changing these to use initializers should fix the build. `std::atomic<bool> stalled{false};`

src/NimBLEL2CAPServer.cpp
```diff
@@ -0,0 +1,35 @@
...
+    for (auto service: this->services) {
+        delete service;
+    }
+}
+
+NimBLEL2CAPChannel* NimBLEL2CAPServer::createService(const uint16_t psm, const uint16_t mtu, NimBLEL2CAPChannelCallbacks* callbacks) {
+
+    auto service = new NimBLEL2CAPChannel(psm, mtu, callbacks);
+
+    if (auto rc = ble_l2cap_create_server(psm, mtu, NimBLEL2CAPChannel::handleL2capEvent, service); rc != 0) {
```
h2zero, line 28: Unfortunately I still need to support C++ 11 so this will need to be made old school:
```
auto rc = ble_l2cap_create_server(psm, mtu, NimBLEL2CAPChannel::handleL2capEvent, service);
if (rc !=0) {
```

src/NimBLEL2CAPChannel.cpp
```diff
@@ -0,0 +1,293 @@
...
+    auto retries = RetryCounter;
+
+    while (retries--) {
+
+        auto txd = os_mbuf_get_pkthdr(&_coc_mbuf_pool, 0);
+        if (!txd) {
+            NIMBLE_LOGE(LOG_TAG, "Can't os_mbuf_get_pkthdr.");
+            return -BLE_HS_ENOMEM;
+        }
+        if (auto res = os_mbuf_append(txd, &(*begin), toSend); res != 0) {
```
h2zero, line 106: Same thing as my comment in L2CAPServer

h2zero (changes requested):
Looks good and thank you for updating this, couple little things to clean up the build.

h2zero:
Thanks @mickeyl, something weird is happening with this PR, it's not triggering the CI or showing the update. Could you squash these commits and force push? Hopefully that triggers the CI.

mickeyl:
@h2zero For some versions it still fails ­– this time because of some bogus use of `min`. I don't think this is inside my code and apparently you're doing some `#undef min` sorcery in your headers… perhaps I need to replicate that?

src/NimBLEL2CAPChannel.cpp
h2zero: Can we move `auto res = os_mbuf_append(txd, &(*begin), toSend);` out of the `if` statement? it causes warnings in some builds.

src/NimBLEL2CAPChannel.h
```diff
@@ -0,0 +1,115 @@
...
+//
+#pragma once
+#ifndef NIMBLEL2CAPCHANNEL_H
+#define NIMBLEL2CAPCHANNEL_H
+
+#include "inttypes.h"
+#include "host/ble_l2cap.h"
+#include "os/os_mbuf.h"
+#include "os/os_mempool.h"
+
```
h2zero, line 12: Unfortunately some nimble headers define `min` and `max` which causes errors when you include stl headers after them. So `#undef min` and `#undef max` will need to be inserted here to fix the build.

h2zero (changes requested):
Looks good, just missed one.

h2zero:
Awesome, thanks @mickeyl

mickeyl:
@h2zero Excellent, glad that this pull request made it. I can now reduce the diff and concentrate on stabilizing the code and probably exposing the (suspected) bugs in the Nimble l2cap infrastructure. Thanks for maintaining this package, I really appreciate that. Expect further pull requests for the l2cap infrastructure coming over the course of this year.

PR #337, opened 2025-05-21
Allow `esp_wifi_remote` >= 0.5.3

johnboiles:
`esp_wifi_remote` >= v0.10.0 is necessary to use esp-nimble-cpp with the latest ESP-IDF master branch.

Alternately, if we can drop `esp_wifi_remote` as a dependency of this project, that'd be even better.

### DIFF
--- a/idf_component.yml
+++ b/idf_component.yml
@@ -16,7 +16,7 @@ dependencies:
     rules:
       - if: "target in [esp32p4]"
   espressif/esp_wifi_remote:
-    version: "<0.5.4"
+    version: ">=0.5.3"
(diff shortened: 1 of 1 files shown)


h2zero:
Thanks @johnboiles, that dependency is only for the P4 and should not affect other MCU builds. 

@chegewara any issues with this change?

chegewara:
It should be safe

johnboiles:
Great! P4 support is evolving rapidly so being able to run on IDF master is really helpful

h2zero:
Thanks!

PR #352, opened 2025-07-31
Adding: EventArgs

yannikHoeflich:
> [!IMPORTANT]
> This PR has some breaking changes and shouldn't be included before Version 3.0

The goal if this change is to give the user a bit more posibilities when it comes to what excactly is send to the other device. But looking forwar into the future, the adding of specific `EventArgs` to the event parameters give the option to easily add more options to the events, without a breaking change. 

By now I included the event args for the `onRead` and `onWrite` events. They give the options to either send custom data back to the device or (in case of `onWrite`) to not save the data that was received. Additionally `onWrite` now gets the old and the new data, so that the user has both.


h2zero:
I like this idea. It doesn't need to be a breaking change, the new callbacks can be added as an overload in addition to the originals.

PR #359, opened 2025-09-05
Updated for device COD, added logging using magic enum and updated for no conflict with ArduinoIoT libs

synapse-2:
Updated for device COD, added logging using magic enum and updated for no conflict with ArduinoIoT libs for the security property enum


h2zero:
This is a bit much, lots of things changed like the readme which are not relevant. Can you break this down a bit and explain the changes and what they are trying to accomplish, such as what the library conflicts are and how you have addressed them?

PR #365, opened 2025-09-10
Find client by handle during disconnect event.

gkoh:
Rather than searching the client list by connection address, search by connection handle.
This avoids an issue where the target device has RPA enabled and addresses will not match.

Additionally, update the client peer address with the ID address when the identity has been resolved.


h2zero:
Awesome, thanks @gkoh

PR #387, opened 2025-11-30
Fix descriptor search range in retrieveDescriptors()

syscode1:
### Problem
The `NimBLERemoteCharacteristic::retrieveDescriptors()` function incorrectly uses the service's end handle (`getRemoteService()->getEndHandle()`) as the search range limit when discovering descriptors.

This causes the following bug:
- Descriptors from **subsequent characteristics** within the same service are incorrectly included
- Results in wrong descriptors being associated with the current characteristic
- Causes memory waste and logical errors

**Example scenario:**
```
Service (Handle 0x0001 - 0x0010)
  ├─ Characteristic A (Handle 0x0002)
  │   ├─ Descriptor A1 (Handle 0x0003)
  │   └─ Descriptor A2 (Handle 0x0004)
  ├─ Characteristic B (Handle 0x0005)
  │   └─ Descriptor B1 (Handle 0x0006)
```

**Before fix:** Searching descriptors for Characteristic A returns A1, A2, **and B1** ❌  
**After fix:** Searching descriptors for Characteristic A returns only A1 and A2 ✅

### Solution
This PR fixes the issue by:
1. Finding the next characteristic in the service
2. Using `next_characteristic_handle - 1` as the end handle for descriptor search
3. Ensuring only descriptors belonging to the current characteristic are retrieved
4. Maintaining backward compatibility for the last characteristic in a service

### Changes Made
- Modified `retrieveDescriptors()` to iterate through service characteristics
- Calculate correct `endHandle` based on the next characteristic's handle
- Falls back to service end handle for the last characteristic
- Added debug logging for the calculated search range

### Testing
This fix correctly handles:
- ✅ Services with multiple characteristics
- ✅ Last characteristic in a service
- ✅ Characteristics with no descriptors

### Files Changed
- `src/NimBLERemoteCharacteristic.cpp` (+18 lines, -2 lines)

### DIFF
--- a/src/NimBLERemoteCharacteristic.cpp
+++ b/src/NimBLERemoteCharacteristic.cpp
@@ -94,8 +94,24 @@ int NimBLERemoteCharacteristic::descriptorDiscCB(
 bool NimBLERemoteCharacteristic::retrieveDescriptors(NimBLEDescriptorFilter* pFilter) const {
     NIMBLE_LOGD(LOG_TAG, ">> retrieveDescriptors() for characteristic: %s", getUUID().toString().c_str());
 
+    const auto pSvc      = getRemoteService();
+    uint16_t   endHandle = pSvc->getEndHandle();
+
+    // Find the handle of the next characteristic to limit the descriptor search range.
+    const auto& chars = pSvc->getCharacteristics(false);
+    for (auto it = chars.begin(); it != chars.end(); ++it) {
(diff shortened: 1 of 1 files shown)


h2zero:
Thanks! Looks like I broke this in 2.0.0 and hasn't been noticed until now. Could you please rebase this so only the changes required are in the PR?

Issue #374, opened 2025-09-27
missing initialize for member 'ble_gap_desc_params::disable_observer_mode' in NimBLEScan.cpp:37:7


chmorgan:
```
[1463/1493] Building CXX object esp-idf/esp-nimble-cpp/CMakeFiles/__idf_esp-nimble-cpp.dir/src/NimBLEScan.cpp.obj
/firmware/components/esp-nimble-cpp/src/NimBLEScan.cpp: In constructor 'NimBLEScan::NimBLEScan()':
//firmware/components/esp-nimble-cpp/src/NimBLEScan.cpp:36:7: warning: missing initializer for member 'ble_gap_disc_params::disable_observer_mode' [-Wmissing-field-initializers]
   36 |       m_scanParams{0, 0, BLE_HCI_SCAN_FILT_NO_WL, 0, 1, 1},
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[1490/1493] Generating ld/sections.ld
```

h2zero:
I saw this too, upstream changes in newer versions :(, easy fix though and only a warning.

PR #400, opened 2026-03-12
Use designated initializer for NimBLEScan parameters and conditionally set disable_observer_mode

stdstijn:
Use designated initializer for NimBLEScan::m_scanParams and guard disable_observer_mode field for ESP-IDF >= 5.4.2

The NimBLEScan constructor previously used positional struct initialization, which no longer matches the ble_gap_disc_params layout in newer ESP-IDF versions (>= 5.4.2) where the field `disable_observer_mode` was added.

This is caused by -Wmissing-field-initializers.

Switch to designated initializers to make the field assignments explicit and more robust across ESP-IDF/NimBLE revisions. The new field is only initialized when building against ESP-IDF 5.4.2 or later to maintain backwards compatibility.


stdstijn:
Proposed fix for issue #374

stdstijn:
Done, if I understand your suggestion correctly.

PR #421, opened 2026-04-24
Replace task noitifcations with semaphores to avoid conflicts.

h2zero:
This will prevent a case where applications may notify a task that is blocking on a BLE operation before it completes. 
This can occur when writing to a characteristic in the application and blocking with task notifications until a notification occurs that releases it and the notification arrives before the write response.

## Summary by CodeRabbit

* **Chores**
  * Removed a deprecated FreeRTOS configuration option.

* **Bug Fixes**
  * Ensures full resource cleanup during device deinitialization.
  * Improves task synchronization and error logging for more reliable operation.

* **Refactor**
  * Internal synchronization reworked to a centralized semaphore registry for cleaner lifecycle and teardown management.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -197,11 +197,4 @@ config NIMBLE_CPP_DEBUG_ASSERT_ENABLED
         Enabling this option will add debug asserts to the NimBLE CPP library.
         This will use approximately 1kB of flash memory.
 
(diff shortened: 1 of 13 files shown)


h2zero:
@coderabbitai review

Issue #423, opened 2026-05-26
Doc: `NimBLEAddress(const uint8_t[6], uint8_t)` silently reverses bytes; "native ESP representation" is ambiguous


fl4p:
### What's going on

`NimBLEAddress::NimBLEAddress(const uint8_t address[BLE_DEV_ADDR_LEN], uint8_t type)` calls `std::reverse_copy` on the input before storing it in `val[]`:

```cpp
NimBLEAddress::NimBLEAddress(const uint8_t address[BLE_DEV_ADDR_LEN], uint8_t type) {
    std::reverse_copy(address, address + BLE_DEV_ADDR_LEN, this->val);
    this->type = type;
}
```

The accompanying docstring says:

> Constructor for compatibility with bluedroid esp library using native ESP representation.

This is technically correct (Bluedroid's `esp_bd_addr_t` is MSB-first; NimBLE's `ble_addr_t.val` is LSB-first; the reverse bridges them), but "native ESP representation" is ambiguous because both stacks are "native" on ESP. A reader who has a `uint8_t[6]` from any other source — a parsed scan, an HCI dump, an over-the-wire address from a higher-level protocol — has no signal whether to pre-reverse it.

The neighbouring `NimBLEAddress(const uint64_t&, uint8_t)` constructor is well-documented:

> Use the same byte order, so use 0xa4c1385def16 for "a4:c1:38:5d:ef:16"

…which leaves the byte-array variant as the only address constructor whose expected byte order isn't stated.

### How it bites in practice

In a project bridging a host-side protocol (aioesphomeapi) to NimBLE, I had MAC `20:A1:11:02:23:45` as a `uint64_t = 0x20a111022345`. I wrote it to a `uint8_t[6]` in LSB-first wire order (`{0x45, 0x23, 0x02, 0x11, 0xa1, 0x20}`) — what NimBLE callbacks would have given me natively — and passed it to `NimBLEAddress(bytes, type)`.

`reverse_copy` then stored `val = {0x20, 0xa1, 0x11, 0x02, 0x23, 0x45}` (MSB-first in the LSB-first slot). NimBLE used that directly in `ble_gap_connect`, sending CONNECT_REQ for a nonexistent peer. The connect timed out with `BLE_HS_ETIMEOUT` (reason 13) — looked indistinguishable from "out of range," because NimBLE was scanning for an address that simply never appears on-air.

The bug masked itself well:

[...]

PR #424, opened 2026-05-26
Doc: clarify NimBLEAddress(uint8_t[6], type) expects MSB-first bytes

fl4p:
Documentation-only follow-up to #423.

The byte-array `NimBLEAddress` constructor calls `std::reverse_copy` on its input — fine for callers passing Bluedroid-style MSB-first bytes (`esp_bd_addr_t`), but easy to miss for anyone arriving with LSB-first bytes (e.g. straight out of a NimBLE `ble_addr_t.val`) since both layouts are "native ESP." The original docstring said:

> Constructor for compatibility with bluedroid esp library using native ESP representation.

Updated docstring spells out the expected order, contrasts it with `NimBLEAddress(ble_addr_t)` for LSB-first wire bytes, and points hex-literal users at `NimBLEAddress(uint64_t, uint8_t)` (which is already well-documented).

No behaviour change.

## How this came up

Tracked down a 100% connect-timeout bug in a NimBLE bridge where I had a `uint64_t` MAC, converted to a `uint8_t[6]` in LSB-first wire order, and passed that to this constructor. `reverse_copy` then stored it MSB-first in `val[]`, NimBLE used `val[]` directly as the on-air LE bytes, and every `ble_gap_connect` timed out scanning for a peer that doesn't exist. Symptom was indistinguishable from "out of range." Switching to the `uint64_t` constructor fixed it immediately. Issue #423 has the full diagnostic story.

## Summary by CodeRabbit

* **Documentation**
  * Improved BLE address constructor documentation with enhanced clarity on byte order handling, parameter descriptions, and guidance on alternative constructors for different input formats.

[Review Change Stack](https://app.coderabbit.ai/change-stack/h2zero/esp-nimble-cpp/pull/424?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/NimBLEAddress.cpp
+++ b/src/NimBLEAddress.cpp
@@ -97,8 +97,19 @@ NimBLEAddress::NimBLEAddress(const std::string& addr, uint8_t type) {
 } // NimBLEAddress
 
 /**
- * @brief Constructor for compatibility with bluedroid esp library using native ESP representation.
- * @param [in] address A uint8_t[6] or esp_bd_addr_t containing the address.
+ * @brief Constructor accepting a 6-byte MSB-first BLE address (Bluedroid `esp_bd_addr_t` layout).
+ * @details The input bytes are stored internally in LSB-first wire
+ * order (NimBLE's `ble_addr_t.val` layout), so this constructor
+ * **reverses** the input. If you already have LSB-first wire bytes —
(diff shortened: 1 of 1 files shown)


h2zero:
Thanks!

h2zero:
Thanks!

PR #426, opened 2026-06-19
Restore GAP device name after GATT reset

chmorgan:
NOTE: Codex 5.5 created this patch and writeup etc but I reviewed it and it looks like what I would have done by hand, and tested it here and it does resolve the issue and explains why I've been seeing my device misnamed for a bit. Not sure how or when I broke it, or if the issue was library side or a nimble change but figured others might be hitting this issue.

ESP-IDF NimBLE can reset the GAP Device Name when the standard GAP service is reinitialized during GATT server startup. NimBLEDevice::init(deviceName) calls NimBLEDevice::setDeviceName(deviceName), but that happens before the server is started. Later, starting advertising starts the server and resets/re-registers the GAP service, causing the C GAP name backing storage to be rebuilt from CONFIG_BT_NIMBLE_SVC_GAP_DEVICE_NAME. In projects where that Kconfig default remains "nimble", the advertised Local Name can be correct while the Generic Access Device Name characteristic reports "nimble".

The failing sequence is:

  -> pAdvertising->start()
     -> NimBLEAdvertising::start()
        -> NimBLEDevice::getServer()->start()
           -> NimBLEServer::resetGATT()
              -> ble_gatts_reset()
              -> ble_svc_gap_init()
                 -> ESP-IDF NimBLE resets GAP name to CONFIG_BT_NIMBLE_SVC_GAP_DEVICE_NAME
                 -> "nimble"
              -> ble_svc_gatt_init()

Fix this by saving the device name supplied through NimBLEDevice::setDeviceName() and reapplying it immediately after ble_svc_gap_init() in NimBLEServer::resetGATT(). This preserves the C++ wrapper contract that the configured device name remains effective even when the wrapper internally resets the GATT/GAP service tables.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the Bluetooth device name and appearance when the GATT server is reset, so these settings no longer revert unexpectedly after a reset.

### DIFF
--- a/src/NimBLEServer.cpp
+++ b/src/NimBLEServer.cpp
@@ -889,8 +889,21 @@ bool NimBLEServer::resetGATT() {
     NimBLEDevice::stopAdvertising();
 # endif
 
+    // esp-idf nimble clears the device name and
+    // appearance during ble_gatts_reset()
+#ifndef CONFIG_USING_NIMBLE_COMPONENT
+    std::string name(ble_svc_gap_device_name());
+    uint16_t appearance = ble_svc_gap_device_appearance();
+#endif
+
     ble_gatts_reset();
     ble_svc_gap_init();
+
+#ifndef CONFIG_USING_NIMBLE_COMPONENT
+    ble_svc_gap_device_name_set(name.c_str());
+    ble_svc_gap_device_appearance_set(appearance);
+#endif
+
     ble_svc_gatt_init();
 
     for (auto svcIt = m_svcVec.begin(); svcIt != m_svcVec.end();) {


h2zero:
Looks like espressif changed the behavior of this :(
https://github.com/espressif/esp-nimble/blob/685675c0128deafdd201c9eb82e61d227364646c/nimble/host/services/gap/src/ble_svc_gap.c#L597

I think a better solution is to save the values in `resetGATT` and reapply the name and appearance after reinitializing the gap services.

Personally I don't use espressifs fork anymore for this reason.

src/NimBLEServer.cpp
```diff
@@ -891,6 +891,10 @@ bool NimBLEServer::resetGATT() {
```
h2zero, line 891: Would be better if we just added these lines here:
```
std::string name(ble_svc_gap_device_name());
uint16_t appearance = ble_svc_gap_device_appearance();
```

src/NimBLEServer.cpp
```diff
@@ -891,6 +891,10 @@ bool NimBLEServer::resetGATT() {
 
     ble_gatts_reset();
     ble_svc_gap_init();
```
h2zero, line 900: Follwed up with:
```
ble_svc_gap_device_name_set(name.c_str());
ble_svc_gap_device_appearance_set(appearance);
```
Then we don't need to add any static members.

h2zero (changes requested):

h2zero:
@finger563 

If your confusion was my remark about the espressif fork, I use this instead: https://github.com/h2zero/esp-nimble-component

Updated to mynewt/nimble 1.9, doesn't have much of the espressif changes and is more stable.

chmorgan:
Ok if I fix it with the espressif version? Should we detect which stack and
only apply it for espressif version?

h2zero:
Sure, we can use this flag: `#ifndef CONFIG_USING_NIMBLE_COMPONENT`

finger563:
Has that been updated or does it work for esp idf v6?

h2zero:
It's a bit of work in progress still, haven't checked it against IDF v6. The same code is being used in nimble-arduino as well so it will be well tested.

h2zero:
@finger563, updated now, idf 5 and 6 working.

chmorgan:
> Looks like espressif changed the behavior of this :( https://github.com/espressif/esp-nimble/blob/685675c0128deafdd201c9eb82e61d227364646c/nimble/host/services/gap/src/ble_svc_gap.c#L597
> 
> I think a better solution is to save the values in `resetGATT` and reapply the name and appearance after reinitializing the gap services.
> 
> Personally I don't use espressifs fork anymore for this reason.

are you using that mynewt version with esp-idf or another platform? I figured there were a ton of porting changes present to support the particular ble devices so it seemed overwhelming to consider going that route.

h2zero:
It's working with IDF, took a bit of work for sure but no too bad

Issue #428, opened 2026-07-01
Allow an application to reject a write from onWrite (return an ATT error to the client)


Nicba1010:
# Allow an application to reject a write from `onWrite` (return an ATT error to the client)

Right now there's no way for a `NimBLECharacteristicCallbacks::onWrite()` handler to reject a write and have the peer actually find out about it. `onWrite()` returns `void`, and the GATT access handler hardcodes a success return on the write path, so no matter what the application does inside the callback, the client always gets an acknowledgement that the write succeeded.

The underlying C stack is perfectly happy to send an ATT error here — the wrapper just doesn't give application code a way to ask for one. I'd like to close that gap.

Checked against `master` @ `1eddc28515bb` (also present in the 2.5.0 release).

## Where it gets dropped

`NimBLEServer::handleGattEvent` is the `ble_gatt_access_fn` registered with the host. Its read path forwards an ATT error when the wrapper itself can't satisfy the request, and the length guards reject oversize writes — but the write path throws away whatever the callback did:

```cpp
// src/NimBLEServer.cpp, handleGattEvent()
return rc == 0 ? 0 : BLE_ATT_ERR_INSUFFICIENT_RES;          // read path forwards an error
if (len > maxLen) return BLE_ATT_ERR_INVALID_ATTR_VALUE_LEN; // length is rejected for us

pAtt->writeEvent(buf, len, peerInfo);   // application callback runs here...
return 0;                               // ...and its outcome is discarded
```

`writeEvent()` and the callback it calls are both `void`, all the way down to the base attribute:

```cpp
// src/NimBLECharacteristic.cpp
void NimBLECharacteristic::writeEvent(const uint8_t* val, uint16_t len, NimBLEConnInfo& connInfo) {
    setValue(val, len);
    m_pCallbacks->onWrite(this, connInfo);   // nothing can come back out
}

// src/NimBLECharacteristic.h
virtual void onWrite(NimBLECharacteristic*, NimBLEConnInfo&);

// src/NimBLELocalValueAttribute.h
virtual void writeEvent(const uint8_t*, uint16_t, NimBLEConnInfo&) = 0;
```

So the wrapper surfaces ATT errors it generates itself, but never lets the application generate one on a write.

## Is this even supported by the underlying stack? (yes — with one caveat)

When traced through the NimBLE host, a non-zero return from the access callback turns out to be honored — so this isn't proposing something the stack would just ignore:

[...]

PR #429, opened 2026-07-01
Add onReadStatus/onWriteStatus callbacks so an application can reject a read/write with an ATT error

Nicba1010:
# Add onReadStatus/onWriteStatus callbacks so an application can reject a read/write with an ATT error

Addresses #428.

Short version of the problem (full detail in the issue): `onWrite()` returns `void`
and `handleGattEvent` hardcodes a success return on the write path, so there's no way
for an application to reject a write and have the peer receive an ATT error — even under
write-with-response, where the protocol allows it. The C host honors a non-zero return
from the access callback; the wrapper just doesn't expose that to application code.

## What this does

Adds two opt-in, status-returning callbacks to `NimBLECharacteristicCallbacks`:

```cpp
virtual int onReadStatus (NimBLECharacteristic*, NimBLEConnInfo&);
virtual int onWriteStatus(NimBLECharacteristic*, NimBLEConnInfo&);
```

Each returns `0` to accept or a `BLE_ATT_ERR_*` code to reject. Their default
implementations call the existing `onRead`/`onWrite` and return `0`, so behavior is
unchanged for anyone who doesn't override them. Internally the `readEvent`/`writeEvent`
chain (`NimBLELocalValueAttribute` → `NimBLECharacteristic`/`NimBLEDescriptor`) now
returns `int`, and `handleGattEvent` forwards that code to the host instead of always
returning `0`. On a rejected read it returns before appending the value.

Six files, +51/-15. No public signature changes to existing methods.

## Backward compatibility

Nothing existing changes shape. `onRead`/`onWrite` keep their `void` signatures, and
every current override keeps working untouched — the new `onReadStatus`/`onWriteStatus`
default to invoking them and accepting. It's purely additive.

I went with the additive form because it drops into a minor release without touching
anyone. That said — since master is already `3.0.0-dev`, if you'd rather take the cleaner
breaking version for the major (just change `onRead`/`onWrite` themselves to return `int`,
`0` = accept, and drop the separate status methods), I'm happy to reshape this into that.
It mirrors the C stack one-to-one; the only reason I didn't lead with it is the override
break. Your call on which fits 3.0 better.

## One caveat

A rejection only reaches the peer under write-with-response. A write-without-response
(ATT Write Command) is unacknowledged, so the stack drops the code and sends nothing —
that's per spec, not something this changes. Worth a mention in the docs so nobody expects
a rejection on a `WRITE_NO_RSP` characteristic.

Descriptors are compile-updated to the new `int` return but always accept; the same
`onWriteStatus`-style hook could be added to `NimBLEDescriptorCallbacks` later if there's
demand.

## Testing

Builds clean for `esp32s3`. I also validated it end-to-end on an ESP32-S3 with a macOS
central: a characteristic returning `BLE_ATT_ERR_INSUFFICIENT_RES` from `onWriteStatus`
surfaced on the central as `CBATTErrorDomain Code=17 "Resources are insufficient."` under
write-with-response, and produced no error under write-without-response — matching the
caveat above. Existing writes to unmodified characteristics were unaffected.

There's an open question from the issue worth settling here too: `writeEvent` commits
`setValue()` before the callback runs, so on a rejected write the local value has already
been updated. This PR keeps that ordering; if you'd prefer the pending `(data, len)` be
passed to `onWriteStatus` so an app can validate before commit, I can add that.

---

### AI usage disclosure

For transparency: I used an AI assistant while working on this — to help trace the call
path through the C host stack, cross-check the behavior against the source, and draft the
code. Flagging that so it's on the table. The changes are small and the cited code paths
point at specific locations, so they're straightforward to review against the tree.

## Summary by CodeRabbit

* **New Features**
  * Read and write requests can now return ATT status codes to accept or reject operations.
  * Added characteristic callback hooks for read/write status codes (with default behavior that preserves existing callbacks).
* **Bug Fixes**
  * GATT handling now properly propagates application accept/reject results for reads and writes.
  * Descriptor and characteristic read/write handling now follows the status-based behavior, including reverting characteristic value changes when a write is rejected.

### DIFF
--- a/src/NimBLECharacteristic.cpp
+++ b/src/NimBLECharacteristic.cpp
@@ -433,8 +433,16 @@ int NimBLECharacteristic::readEvent(NimBLEConnInfo& connInfo) {
  * @param [in] connInfo A reference to a NimBLEConnInfo instance containing the peer info.
  */
 int NimBLECharacteristic::writeEvent(const uint8_t* val, uint16_t len, NimBLEConnInfo& connInfo) {
+    // Commit before the callback so getValue() reflects the new value inside it
+    // (onWriteStatus defaults to calling the legacy onWrite), but snapshot first
+    // so a rejected write can be rolled back and never becomes observable state.
+    NimBLEAttValue previous = m_value;
     setValue(val, len);
-    return m_pCallbacks->onWriteStatus(this, connInfo);
+    const int rc = m_pCallbacks->onWriteStatus(this, connInfo);
+    if (rc != 0) {
+        m_value = previous;
+    }
+    return rc;
 } // writeEvent
 
 /**
--- a/src/NimBLECharacteristic.cpp
+++ b/src/NimBLECharacteristic.cpp
@@ -422,8 +422,8 @@ void NimBLECharacteristic::updatePeerStatus(const NimBLEConnInfo& peerInfo) cons
  * @brief Handle a read event from a client.
  * @param [in] connInfo A reference to a NimBLEConnInfo instance containing the peer info.
  */
-void NimBLECharacteristic::readEvent(NimBLEConnInfo& connInfo) {
-    m_pCallbacks->onRead(this, connInfo);
+int NimBLECharacteristic::readEvent(NimBLEConnInfo& connInfo) {
+    return m_pCallbacks->onReadStatus(this, connInfo);
 } // readEvent
 
 /**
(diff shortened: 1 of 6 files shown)


h2zero:
Thanks @Nicba1010, I like this change. I'm considering maybe making this a breaking change to avoid adding more callbacks but this is good as well. Maybe we could rename these to be a little more clear as to their function, something like onRead/WriteRequest, what do you think?

Nicba1010:
Thank you for taking a look :). I made it this way so you don't have to do a major release (as is usually with changes as breaking as these) unless you have it planned, but I'm fine both ways. Regarding the naming, anything is fine to me, I'm not particularly opinionated there :).

PR #430, opened 2026-07-06
NimBLEAttValue.cpp - preserve capacity after realloc failure

chmorgan:
NimBLEAttValue::append advanced m_capacity immediately after calling realloc. If realloc failed, the original buffer remained allocated but the object advertised a larger capacity than it actually owned. A later append could then skip reallocation and copy past the end of the allocation.

Track whether the append actually needs growth and only update m_capacity after the realloc result passes the null check. Appends that fit the existing allocation leave capacity unchanged; failed growth preserves the old capacity invariant.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data append behavior to better handle memory allocation failures.
  * Prevented internal capacity values from being updated when an expansion attempt fails, keeping state consistent and avoiding incorrect size reporting after errors.


h2zero:
LGTM thanks!

PR #439, opened 2026-08-21
Fix NimBLEDevice::init() to check the return value from nimble_port_init().

doudar:
Fix NimBLEDevice::init() to check the return value from nimble_port_init(). The result and npl_funcs pointer are logged for diagnostics, and initialization now returns false on failure instead of continuing with an invalid NimBLE state.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initialization reliability on ESP platforms by detecting and reporting startup failures.
  * Initialization now correctly indicates when setup cannot be completed.

### DIFF
--- a/src/NimBLEDevice.cpp
+++ b/src/NimBLEDevice.cpp
@@ -983,7 +983,15 @@ bool NimBLEDevice::init(const std::string& deviceName) {
 #   endif
 #  endif
 # endif
+# ifdef ESP_PLATFORM
+        esp_err_t rc = nimble_port_init();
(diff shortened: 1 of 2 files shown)


h2zero:
Thanks @doudar, looks good if the issues flagged by the rabbit are fixed. Have you had issues with this failing?

doudar:
Okay, I'll obey the rabbit. 

There were no issues with it failing, but it's a bug/inconsistency I discovered while investigating a connection issue I had.

PR #442, opened 2026-08-30
fix: propagate application write rejections as ATT Error Responses

### LINKED ISSUE #441: GATT server write path cannot propagate application errors to the client (handleGattEvent always returns 0)
## Problem

When an application rejects a write inside `NimBLECharacteristicCallbacks::onWrite` (e.g. device busy, validation failure), the BLE client still receives a **successful** Write Response. There is no way to signal an ATT Error Response for a rejected write.

## Root cause

The NimBLE C stack supports ATT Error Responses from the `ble_gatt_access_fn` callback (its return value is `int`), and a characteristic with the `WRITE` property uses Write Requests (opcode 0x12), which per the Bluetooth spec support error responses. The capability is lost in the C++ wrapper, in three places:

1. `NimBLECharacteristicCallbacks::onWrite` is `void`.
2. `NimBLECharacteristic::writeEvent` is `void`.
3. `NimBLEServer::handleGattEvent` (src/NimBLEServer.cpp, the `OP_WRITE_DSC`/`OP_WRITE_CHR` block) ends with a hardcoded `return 0;` right after `pAtt->writeEvent(buf, len, peerInfo);`.

So business-layer results (which *are* computed synchronously inside the ATT transaction context by applications) are silently discarded, and every write is ACKed as success.

## Constraint

`writeEvent` cannot simply be changed to return `int`: it is a **pure virtual** in `NimBLELocalValueAttribute` (`virtual void writeEvent(...) = 0`) and overridden by both `NimBLECharacteristic` and `NimBLEDescriptor`. Changing its signature is a breaking API change.

## Proposed direction

Keep all signatures unchanged and add a per-write error slot on `NimBLECharacteristic`:

- `void setWriteError(uint8_t attError)` — called by the application inside `onWrite()`; `0` accepts, vendor range `0x80`–`0x9F` for application-defined errors.
- `int getWriteError() const` — read by `handleGattEvent`'s write branch, which returns it instead of `0`.

`writeEvent()` resets the slot to `0` before invoking `onWrite`, and the OP_WRITE_DSC path keeps returning `0` (descriptors carry no slot). Fully backwards compatible: applications that never call `setWriteError` see identical behavior.

## PR

Implemented in #442. Feedback on the design (especially whether you'd prefer a different mechanism, e.g. a new virtual with a default implementation) is welcome.

jie-meng:
## Description

Closes #441.

`handleGattEvent`'s write branch hardcodes `return 0`, so an application that rejects a write inside `onWrite()` (busy, validation failure, ...) can never tell the client — the client always receives a successful Write Response. The NimBLE C stack supports ATT Error Responses from the access callback, and Write Requests (opcode 0x12) support them per spec; the capability is only lost in this wrapper.

Because `NimBLELocalValueAttribute::writeEvent` is a **pure virtual `void`** (overridden by both `NimBLECharacteristic` and `NimBLEDescriptor`), the error cannot be threaded through the existing signatures without a breaking API change. This PR therefore uses a **per-write error slot** on the characteristic — fully backwards compatible:

- `void NimBLECharacteristic::setWriteError(uint8_t attError)` — called by the application inside `onWrite()`; `0` accepts, vendor range `0x80`–`0x9F` for application-defined error codes.
- `int NimBLECharacteristic::getWriteError() const` — read by `handleGattEvent`, which returns the slot value on the `OP_WRITE_CHR` path and `0` on the `OP_WRITE_DSC` path.
- `writeEvent()` resets the slot to `0` before invoking `onWrite()`.

## Changes

| File | Change |
|------|--------|
| `src/NimBLECharacteristic.h` | public `setWriteError` / `getWriteError`, private `int m_writeError = 0` |
| `src/NimBLECharacteristic.cpp` | reset slot in `writeEvent` before `onWrite`; two method implementations |
| `src/NimBLEServer.cpp` | write branch returns the slot via an op-guarded cast (DSC/CHR share one case block; the guard keeps descriptor writes succeeding) |

No existing signature, virtual, or behavior changes for applications that don't call `setWriteError`.

## Client-visible behavior

- Write **Request** (opcode 0x12) with a non-zero slot → the stack sends an **ATT Error Response** with that code. Android surfaces it as a non-`GATT_SUCCESS` status in `onCharacteristicWrite`; iOS as an error in `peripheral:didWriteValueForCharacteristic:error:`.
- Write **Command** (0x52, no response) has no response to carry the error — unchanged (documented in the header comment).

## Known pre-existing behavior (unchanged, for the record)

- `writeEvent` applies `setValue(val, len)` before invoking `onWrite`, so a *rejected* write still updates the locally stored attribute value. This predates this PR; an ATT Error Response tells the client to treat the value as unchanged, and a rollback pass is left out of scope here.

## Validation

- The identical patch, backported to the 2.5.0 branch, is compiled and shipping in a downstream ESP-IDF 5.5.4 / ESP32-S3 product (component-manager build + 727 host-side unit tests, including new tests for the reject/accept paths).
- This branch applies the same patch to `master` (identical code context); it has not been compiled against 3.0.0-dev independently.

## Summary by CodeRabbit

* **New Features**
  * Added write-validation error handling for characteristic callbacks.
  * Applications can now report an ATT error when a characteristic write is rejected.
  * Write errors reset for each new write, preventing stale errors from affecting later requests.
  * Write Commands continue to be accepted without returning an ATT error response.

### DIFF
--- a/src/NimBLECharacteristic.cpp
+++ b/src/NimBLECharacteristic.cpp
@@ -434,8 +434,29 @@ void NimBLECharacteristic::readEvent(NimBLEConnInfo& connInfo) {
  */
 void NimBLECharacteristic::writeEvent(const uint8_t* val, uint16_t len, NimBLEConnInfo& connInfo) {
     setValue(val, len);
+    m_writeError = 0;
     m_pCallbacks->onWrite(this, connInfo);
 } // writeEvent
+
+/**
+ * @brief Set the ATT error code to return for the current write operation.
+ * @param [in] attError The ATT error code; 0 accepts the write. Use the
+ * Bluetooth-spec vendor-specific range 0x80-0x9F for application errors.
+ * Call this from your NimBLECharacteristicCallbacks::onWrite() implementation
+ * to reject the write; the stack will then send an ATT Error Response instead
+ * of a success response. Ignored for Write Commands (no response exists).
+ */
+void NimBLECharacteristic::setWriteError(uint8_t attError) {
+    m_writeError = attError;
+} // setWriteError
+
+/**
+ * @brief Get the write error code set by the onWrite callback.
+ * @return The ATT error code for the current write operation, 0 if none was set.
+ */
+int NimBLECharacteristic::getWriteError() const {
+    return m_writeError;
+} // getWriteError
 
 /**
  * @brief Set the callback handlers for this characteristic.
--- a/src/NimBLECharacteristic.h
+++ b/src/NimBLECharacteristic.h
@@ -76,6 +76,22 @@ class NimBLECharacteristic : public NimBLELocalValueAttribute {
 
     NimBLECharacteristicCallbacks* getCallbacks() const;
 
+    /**
+     * @brief Set the ATT error code to return for the current write operation.
+     * @param [in] attError The ATT error code; 0 accepts the write. Use the
+     * Bluetooth-spec vendor-specific range 0x80-0x9F for application errors.
+     * Call this from your NimBLECharacteristicCallbacks::onWrite() implementation
+     * to reject the write; the stack will then send an ATT Error Response instead
+     * of a success response. Ignored for Write Commands (no response exists).
+     */
+    void setWriteError(uint8_t attError);
+
+    /**
+     * @brief Get the write error code set by the onWrite callback.
+     * @return The ATT error code for the current write operation, 0 if none was set.
+     */
+    int getWriteError() const;
+
     /*********************** Template Functions ************************/
 
 # if __cplusplus < 201703L
@@ -305,6 +321,7 @@ class NimBLECharacteristic : public NimBLELocalValueAttribute {
     NimBLEService*                 m_pService{nullptr};
     std::vector<NimBLEDescriptor*> m_vDescriptors{};
     mutable SubPeerArray           m_subPeers{};
+    int                            m_writeError = 0;
 }; // NimBLECharacteristic
 
 /**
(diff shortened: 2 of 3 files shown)


jie-meng:
One question on release strategy: would you also want this backported to the `release/2.5` branch (e.g. as part of a 2.5.1 patch release)? Asking because a downstream project of ours is currently pinned to the 2.5.x line and is consuming this fix from a fork in the meantime — knowing your backport plans would tell us when we can switch back to the upstream package. No rush either way; happy to send a release/2.5 PR if you'd prefer that.

jie-meng:
@coderabbitai The Docstring Coverage check reports 62.50% (5/8 functions documented). Could you list exactly which 3 functions are counted as missing docstrings, so I can add them? Thanks!

jie-meng:
Thanks for the analysis — agreed, treating the Docstring Coverage warning as a false positive. Every touched declaration/definition in this diff (setWriteError/getWriteError in both header and implementation, writeEvent, handleGattEvent) already carries a preceding Doxygen block, so I'll leave the code as is rather than add speculative docstrings.

jie-meng:
@h2zero - ping on this PR. We've been using this fix in production via our fork and it's working well. Would appreciate a review/merge when you have cycles. Also opened backport #444 for release/2.5.

h2zero:
Thanks @jie-meng, I'm leaning toward #429 to address this.
Ideally the callbacks should just return an error code there instead, but that will be a breaking change.

